PR #8, opened 2020-09-10
Add golangci linter and lint the code accordingly

rporres:
Added golangci linter with a hive inspired configuration. We may want to tune it in the future to suit our needs better.

120 lines is an arbitrary line length that should be long enough to accommodate all styles while avoiding line continuations


pkg/validations/requests_limits.go
```diff
@@ -60,18 +61,19 @@ func (r *RequestLimitValidation) Validate(request reconcile.Request, kind string
 		return
 	}
 
-	replica_cnt := reflect.ValueOf(obj).FieldByName("Spec").FieldByName("Replicas").Elem().Int()
-	if replica_cnt > 0 {
-		podTemplateSpec := reflect.ValueOf(obj).FieldByName("Spec").FieldByName("Template").Interface().(v1.PodTemplateSpec)
+	replicaCnt := reflect.ValueOf(obj).FieldByName("Spec").FieldByName("Replicas").Elem().Int()
+	if replicaCnt > 0 {
+		podTemplateSpec := reflect.ValueOf(obj).FieldByName("Spec").
```
rrati, line 66: I'm not a big fan of doing this kind of thing in Go.  It makes the line feel disjointed to me and not as connected as a stream of function calls.  Similar to the calls with args (like the log lines).  Is this a new-ish standard in the Go world?

rporres: It is just a way to avoid enormous lines. I can try to format it in a way that looks clearer if you don't like it

rporres: I've seen this style a lot when chaining multiple function calls to avoid humongous lines
```golang
               podTemplateSpec := reflect.
                       ValueOf(obj).
                       FieldByName("Spec").
                       FieldByName("Template").
                       Interface().
                       (v1.PodTemplateSpec)

rrati: I've seen that style when doing builder type functions as well.  I like that a lot better than the single line break.

rporres: Great, I will do that as well with the function calls

rporres: done in 210003d

PR #9, opened 2020-09-11
Add vendoring

rporres:
Since this is a standalone program not intended to be consumed as a
library, vendoring makes sense here.


rrati:
Why vendor instead of using go modules?

rporres:
It is still using go modules. And vendoring. Since this is an app and not a library, it is important that we're always able to build the final `manager` command. Hence the `vendor` directory.

jmelis (approved):
LGTM

rrati (approved):

PR #15, opened 2020-09-24
Generate CSV from separated manifest files

rporres:
In this way we can use the manifests to install the operator without the
whole OLM infrastructure.


rrati (approved):
LGTM

PR #16, opened 2020-09-24
Rename repo to deployment-validation-operator

rporres:
While dv-operator is shorter, it does not tell anything about the
purpose of the operator. The convention of openshift operators is to
have long and descriptive names, so let's follow it.


rporres:
Once we merge this we should rename the repository

rrati (approved):
LGTM

PR #18, opened 2020-09-25
Prepare for CI

rporres:
* Image is built through a multi-stage build to avoid depending on the
  go interpreter from the environment
* Create a make target to test using a Docker to avoid depending on the
  go interpreter from the environment
* Added a make target to login and push


rporres:
We will revisit the image creation when we have something defined for `opm`

rrati (approved):
LGTM

PR #20, opened 2020-09-25
Add new target in Makefile for docker login

rrati:
This is needed because docker needs to be logged into before generating
the image in order for the proper username to be included in the docker
image name


mmclanerh:
lgtm!

mmclanerh (approved):
lgtm

rporres:
The config part was the reason why I had used just one target as the login config won't be visible for the more general docker-push target

rporres:
ok, I see you fixed it in https://github.com/app-sre/deployment-validation-operator/pull/21 :-D

PR #21, opened 2020-09-25
Removed additional docker login make target and set user in build_deploy

rrati:
When creating an image, the org used when generating the image
name is defined by the USER env which is usually the username
running the command.  This isn't always the desired namespace for
things like CI/CD

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-make docker-login-and-push
+USER=$QUAY_USER make docker-login-and-push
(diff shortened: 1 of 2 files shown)


mmclanerh (approved):

mmclanerh:
lgtm glhf!

build_deploy.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-make docker-login-and-push
+USER=$QUAY_USER make docker-login-and-push
```
rporres, line 3: The only place where this is used in our repo is setting the `IMAGE_REPOSITORY` in `boilerplate/openshift/golang-osd-operator/project.mk`. Why don't we set that directly, this is quite misleading

PR #22, opened 2020-09-25
Fix errant make dependency for docker-login-and-push

rrati:
Forgot to remove the docker-login make target dependency


mmclanerh (approved):
lgtm! fixes the adjusted fixes. :D

PR #26, opened 2020-09-28
Image pushes to quay.io/app-sre

rporres:
We may want in the future to copy it to a repository that is more
community oriented, but for the moment this should make things work
properly with app-sre automation


rrati:
We should make the user/org something that is passed into the template, defaulting to the upstream location.

rporres:
I would refactor then the whole bundle Makefile to have everything in a single location... Created a card for it https://issues.redhat.com/browse/APPSRE-2530

rrati (approved):
lgtm

PR #28, opened 2020-09-28
Make REPLACE_VERSION optional when generating the CSV for the bundle.

rrati:
The first time a bundle is built there will be no version to replace,
so need to make it optional and remove references to it in the CSV.

### DIFF
--- a/deploy/bundle/Makefile
+++ b/deploy/bundle/Makefile
@@ -24,6 +24,9 @@ manifest:
 	TEMPLATE=`mktemp` ; \
 	./generate-csv-template.py > $${TEMPLATE} ; \
 	oc process --local -o yaml --raw=true IMAGE=$(OPERATOR_IMAGE) IMAGE_TAG=$(OPERATOR_IMAGE_TAG) VERSION=$(VERSION) REPLACE_VERSION=$(REPLACE_VERSION) -f $${TEMPLATE} > $(CSV)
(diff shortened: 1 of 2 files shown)


deploy/bundle/Makefile
```diff
@@ -24,6 +24,9 @@ manifest:
 	TEMPLATE=`mktemp` ; \
 	./generate-csv-template.py > $${TEMPLATE} ; \
 	oc process --local -o yaml --raw=true IMAGE=$(OPERATOR_IMAGE) IMAGE_TAG=$(OPERATOR_IMAGE_TAG) VERSION=$(VERSION) REPLACE_VERSION=$(REPLACE_VERSION) -f $${TEMPLATE} > $(CSV)
```
rporres, line 26: missing a `"\;"` at the end

deploy/bundle/Makefile
```diff
@@ -24,6 +24,9 @@ manifest:
...
+	@if [ "${REPLACE_VERSION}" == "" ]; then \
+	  sed -i.bak "/replaces/d" $(CSV); \
```
rporres, line 28: Why do you need to create a copy? you can do an in-place edit using `-i ""`

rrati: on Mac, the -i REQUIRES a copy to be made.

rrati: Actually, which we'll need to delete.  *sigh*

PR #29, opened 2020-09-28
Remove the backup csv generated when removing replaces from the manifest

rrati:
If this isn't removed there are effectively 2 csvs in the manifest
dir, which isn't good


deploy/bundle/Makefile
```diff
@@ -26,6 +26,7 @@ manifest:
 	oc process --local -o yaml --raw=true IMAGE=$(OPERATOR_IMAGE) IMAGE_TAG=$(OPERATOR_IMAGE_TAG) VERSION=$(VERSION) REPLACE_VERSION=$(REPLACE_VERSION) -f $${TEMPLATE} > $(CSV)
 	@if [ "${REPLACE_VERSION}" == "" ]; then \
 	  sed -i.bak "/replaces/d" $(CSV); \
+	  rm -f $(CSV).bak; \
```
rporres, line 29: Why don't you just do an inplace in the previous line?

```shell
sed -i "" -e "/replaces/d" $(CSV)
```

rrati: Yep, that's better

rporres: nah, that won't work in Linux properly :(

rporres (approved):

PR #30, opened 2020-09-28
Added script to build manifests and catalogs

rrati:


build_catalog.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+num_commits=$(git rev-list $(git rev-list --max-parents=0 HEAD)..HEAD --count)
+current_commit=$(git rev-parse --short=7 HEAD)
+version="0.1.$num_commits-$current_commit"
```
rporres, line 5: we can get this from the boilerplate vars: https://github.com/app-sre/deployment-validation-operator/blob/ea2120e886208c1a24a13171d1f8ed10320f5224/boilerplate/openshift/golang-osd-operator/standard.mk#L22-L24

build_catalog.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Login to docker
+docker_conf="$PWD/.docker"
+mkdir -p "$docker_conf"
+docker_cmd="docker --config=$docker_conf"
+
+$docker_cmd login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
+
+# Find the CSV version from the previous bundle
+$docker_cmd pull $BUNDLE_IMAGE_NAME:latest && exists=1 || exists=0
```
rporres, line 16: if you're depending from external variables you should check for them to exist or at least add `set -o no unset`

build_catalog.sh
```diff
@@ -0,0 +1,49 @@
...
+    docker export tmp_$$ | tar -xf -
+    prev_version=`find . -name *.clusterserviceversion.* | xargs cat - | yq r - metadata.name`
+  popd
+  rm -rf $tmp_dir
+  docker rm tmp_$$
+fi
+
+# Build/push the new bundle
+pushd deploy/bundle
+  IMAGE=$BUNDLE_IMAGE_NAME IMAGE_TAG=$current_commit VERSION=$version REPLACE_VERSION=$prev_version make bundle
```
rporres, line 31: Does this work ok if `prev_version' is empty? If so, I would make it more explicit, as it can be confusing

build_catalog.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+num_commits=$(git rev-list $(git rev-list --max-parents=0 HEAD)..HEAD --count)
+current_commit=$(git rev-parse --short=7 HEAD)
+version="0.1.$num_commits-$current_commit"
+opm_version="1.14.0"
```
rporres, line 6: we may this to be a var that can be set externally with a default

build_catalog.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Login to docker
+docker_conf="$PWD/.docker"
+mkdir -p "$docker_conf"
+docker_cmd="docker --config=$docker_conf"
```
rporres, line 11: the boilerplate uses the `CONTAINER_ENGINE` instead of docker for those using podman

rporres (changes requested):
Looks great. I love how we don't have to depend on keeping the previous version anywhere

A couple of things:

* I would integrate this with the Makefile to avoid duplicating logic.
* I would also make it a bit more defensive, with a `set -eu` at the beginning

build_catalog.sh
rrati: It does.  That was the sed logic we were trying to reshape to work across linux and mac

build_catalog.sh
rrati: Will do this once things are working and can integrate with makefile/boilerplate.  Too many variables atm

build_catalog.sh
rrati: Checking for the env names.  Can't use -o nounset because there is a variable that will be unbound in some cases (previous version)

build_catalog.sh
rrati: I thought about that, but I'm not sure we want a bunch of different opm versions in use.  For now I want to keep as is and we can iterate to determine how to handle the version.

build_catalog.sh
rrati: Will use once this gets integrated with boilerplate.

build_catalog.sh
rporres: fair enough

build_catalog.sh
rporres: ok

build_catalog.sh
rporres: ok

build_catalog.sh
```diff
@@ -0,0 +1,62 @@
...
+if [ -z "${BUNDLE_IMAGE_NAME}" ]; then
+  echo "BUNDLE_IMAGE_NAME is not set"
+  exit 1
+fi
+
+if [ -z "${CATALOG_IMAGE_NAME}" ]; then
+  echo "CATALOG_IMAGE_NAME is not set"
+  exit 1
+fi
+
```
rporres, line 12: I usually do this in the following way to capture everything in one go

```bash
count=0
for var in BUNDLE_IMAGE_NAME \
           CATALOG_IMAGE_NAME \
           QUAY_USER \
           QUAY_TOKEN
do
    if [ ! "${!var}" ]; then
        echo "$var is not set"
        count=$((count + 1))
    fi
done

[ $count -gt 0 ] && exit 1
```

rporres (approved):
just a small suggestion. Looks good otherwise

build_catalog.sh
rrati: I didn't include QUAY_USER/TOKEN simply because they are part of the CI/CD, but you're right they should be included.  I added ^^.

PR #31, opened 2020-09-29
Add ability to ignore certain namespaces

rporres:
This is very handy to avoid generating metrics we're not going to use... I don't know if we were planning to do something fancier with config. This is nice and easy and it is exactly what we need 😂 

### DIFF
--- a/pkg/controller/generic_reconciler.go
+++ b/pkg/controller/generic_reconciler.go
@@ -3,7 +3,9 @@ package controller
 import (
 	"context"
 	"fmt"
+	"os"
 	"reflect"
+	"regexp"
 	"strings"
 
 	"github.com/app-sre/deployment-validation-operator/pkg/validations"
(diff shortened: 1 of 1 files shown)


pkg/controller/generic_reconciler.go
```diff
@@ -63,6 +73,11 @@ func (gr *GenericReconciler) Reconcile(request reconcile.Request) (reconcile.Res
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.V(2).Info("Reconcile", "Kind", gr.reconciledKind)
 
+	if namespaceIgnore != nil && namespaceIgnore.Match([]byte(request.Namespace)) {
+		reqLogger.Info("Ignoring object as it matches namespace ignore pattern")
```
rrati, line 77: It would be helpful to print the namespace/name/kind of the object.  It's always possible the regex is wrong and is excluding objects that are actually desired to be watched.  Proving logging as to what is excluded will help debug.

rporres: This is how it gets printed currently: `{"level":"info","ts":1601385823.4955745,"logger":"DeploymentController","msg":"Ignoring object as it matches namespace ignore pattern","Request.Namespace":"openshift-apiserver-operator","Request.Name":"openshift-apiserver-operator"}`

rrati: Oh, right.  Duh.

rrati (approved):

PR #33, opened 2020-09-30
Remove Operator Metrics Server

rporres:
It is not serving any metric, so unless it is used for something I don't understand I would remove it and just have one single metrics endpoint.


rrati (approved):
LGTM

PR #34, opened 2020-10-02
Fixes to catalog/bundle generation

rrati:
The bundle was being generated with the incorrect operator image
tag which causes a failure in catalog generation.  Since latest
tag is used that tag can not be pushed unless bundle and catalog
generation succeed.  Also changed the bundle/catalog env names to
be more consistent with other names for images

### DIFF
--- a/build_catalog.sh
+++ b/build_catalog.sh
@@ -1,13 +1,19 @@
 #!/bin/bash
+set -e
+
+function log ()
+{
+  echo "######## $1 ########"
+}
 
 count=0
-for var in BUNDLE_IMAGE_NAME \
-           CATALOG_IMAGE_NAME \
+for var in BUNDLE_IMAGE \
+           CATALOG_IMAGE \
            QUAY_USER \
            QUAY_TOKEN
 do
   if [ ! "${!var}" ]; then
-    echo "$var is not set"
+    log "$var is not set"
     count=$((count + 1))
   fi
 done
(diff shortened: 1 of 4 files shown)


build_catalog.sh
```diff
@@ -27,57 +33,73 @@ docker_cmd="docker --config=$docker_conf"
 $docker_cmd login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
 
 # Find the CSV version from the previous bundle
-echo "pulling laster bundle image"
-$docker_cmd pull $BUNDLE_IMAGE_NAME:latest && exists=1 || exists=0
+log "Pulling lastest bundle image $BUNDLE_IMAGE"
```
rporres, line 36: s/lastest/latest/

build_catalog.sh
```diff
@@ -27,57 +33,73 @@ docker_cmd="docker --config=$docker_conf"
...
 pushd deploy/bundle
-  echo "creating the bundle"
+  log "Creating bundle $BUNDLE_IMAGE:$current_commit"
   if [[ $prev_version != "" ]]; then
     export REPLACE_VERSION=$prev_version
   fi
-  export IMAGE=$BUNDLE_IMAGE_NAME
-  export IMAGE_TAG=$current_commit
+  export BUNDLE_IMAGE_TAG=$current_commit
+  export OPERATOR_IMAGE_TAG=$current_commit
```
rporres, line 63: shouldn't this be `$version`?

rrati: Yes!  Good catch.

rporres (approved):
Let's make it work and integrate it at some point...

rrati:
Big +1 to that. :)

Don't want to merge this until I have done some other cleanup since we'll need to start from scratch on catalogs.

rporres:
ok, merge at your leisure ;-)

PR #37, opened 2020-10-07
Combine bundle/operator makefiles

rrati:
This allows usage and definition of common variables for operator
build and bundle/catalog builds


Makefile
```diff
@@ -1,7 +1,20 @@
...
 IMAGE_REPOSITORY ?= app-sre
 QUAY_USER ?=
 QUAY_TOKEN ?=
+OPM_VERSION = v1.14.0
+
+BUNDLE_DEPLOY_DIR=deploy/bundle
+MANIFEST_DIR=$(BUNDLE_DEPLOY_DIR)/manifests
+CSV=$(MANIFEST_DIR)/deploymentvalidationoperator.clusterserviceversion.yaml
+OPERATOR_IMAGE=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME)
+OPEARTOR_IMAGE_TAG=v$(OPERATOR_VERSION)
```
rporres, line 13: OPEARTOR? looks like typo...

build_catalog.sh
```diff
@@ -10,7 +10,9 @@ count=0
 for var in BUNDLE_IMAGE \
            CATALOG_IMAGE \
            QUAY_USER \
-           QUAY_TOKEN
+           QUAY_TOKEN \
+           CONTAINER_ENGINE \
+           CONFIG_DIR
```
rporres, line 15: Missing at least (I would double check 😄  )
```
CURRENT_COMMIT
OPERATOR_VERSION
OPM_VERSION
GOOS
GOARCH

File: build_catalog.sh [hunk 3]
```diff
@@ -50,36 +42,36 @@ if [ $exists -eq 1 ]; then
...
-  log "Creating bundle $BUNDLE_IMAGE:$current_commit"
+  log "Creating bundle $BUNDLE_IMAGE:$CURRENT_COMMIT"
   if [[ $prev_version != "" ]]; then
     export REPLACE_VERSION=$prev_version
   fi
-  export BUNDLE_IMAGE_TAG=$current_commit
-  export OPERATOR_IMAGE_TAG=v$version
-  export VERSION=$version
+  export BUNDLE_IMAGE_TAG=$CURRENT_COMMIT
+#  export OPERATOR_IMAGE_TAG=v$version
```
rporres, line 55: remove this line if it is not needed

Makefile
rporres: Suggestion. To avoid redefining what it has already been defined in boilerplate I would move `OPERATOR_IMAGE` and `OPERATOR_IMAGE_TAG` after the include and do it from the `OPERATOR_IMAGE_URI` as follows:

```bash
OPERATOR_IMAGE = $(shell echo ${OPERATOR_IMAGE_URI} | cut -d ":" -f 1)
OPERATOR_IMAGE_TAG = $(shell echo ${OPERATOR_IMAGE_URI} | cut -d ":" -f 2)

Comment (rporres) [reply Makefile:13]: In any case I think you will have to move them anyway since you need things like `IMAGE_REGISTRY` from boilerplate...

File: Makefile [hunk 4]
```diff
@@ -1,7 +1,20 @@
...
+BUNDLE_IMAGE?=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME)-bundle
+CATALOG_IMAGE?=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME)-catalog
+
```
rporres, line 16: Can you define these from `OPERATOR_IMAGE`?

rporres (commented):
Love this! Thanks for the work. A few comments here and there

Makefile
rrati: My intention is to push those definitions into base boilerplate.  All the info is there, it's just in a form that we can't user without extraction logic like you offered.  There's no good reason to need to do that.  Just have more fundamentals in boilerplate and build up the existing vars from those fundamental vars.

rrati: Fixed the typo

rporres: I like the idea of moving those to `boilerplate/openshift/golang-osd-operator/standard.mk` and refactoring how `OPERATOR_IMAGE_URI`, so do it :)

rrati: Pushed the definitions down into boilerplate

Makefile
rrati: Yep.  Good catch.  Done

build_catalog.sh
rrati: Removed

build_catalog.sh
rrati: I added these.  I didn't catch any other missing ones.

Makefile
```diff
@@ -1,7 +1,18 @@
...
 IMAGE_REPOSITORY ?= app-sre
 QUAY_USER ?=
 QUAY_TOKEN ?=
+OPM_VERSION = v1.14.0
+
+BUNDLE_DEPLOY_DIR=deploy/bundle
+MANIFEST_DIR=$(BUNDLE_DEPLOY_DIR)/manifests
+CSV=$(MANIFEST_DIR)/deploymentvalidationoperator.clusterserviceversion.yaml
+BUNDLE_IMAGE?=$(OPERATOR_IMAGE)-bundle
+CATALOG_IMAGE?=$(OPERATOR_IMAGE)-catalog
```
rporres, line 13: I don't understand how this can work if you don't define this two variables after the include... 😅

rrati: Well, it wouldn't. :). Fixed.

rporres (approved):

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -99,7 +103,7 @@ op-generate:
 .PHONY: generate
 generate: op-generate go-generate
 
-# rporresm: Force GOOS=linux as we may want to build containers in darwin
+# upstream-diff: Force GOOS=linux as we may want to build containers in darwin
```
2uasimojo, line 106: We're still in the process of figuring out how to make local targets run in a container -- the *same* container that will be used in appsre and prow (the latter being the bigger challenge). Because even if we set `GOOS=linux` here, I don't think we can rely on the rest of the environment being same-enough to trust the resulting binary.

But in the meantime, I think this would be a reasonable thing to propose upstream.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -20,11 +20,15 @@ CONTAINER_ENGINE=$(shell command -v podman 2>/dev/null || command -v docker 2>/d
...
-# Standard/App-SRE sha length is 7 instead of 8
+# upstream-diff - Standard/App-SRE sha length is 7 instead of 8
 CURRENT_COMMIT=$(shell git rev-parse --short=7 HEAD)
 OPERATOR_VERSION=$(VERSION_MAJOR).$(VERSION_MINOR).$(COMMIT_NUMBER)-$(CURRENT_COMMIT)
 
-IMG?=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME):v$(OPERATOR_VERSION)
+# upstream-diff - set IMG from two sub-components which allows more commonality in use
+OPERATOR_IMAGE=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME)
+OPERATOR_IMAGE_TAG=v$(OPERATOR_VERSION)
+IMG?=$(OPERATOR_IMAGE):$(OPERATOR_IMAGE_TAG)
```
2uasimojo, line 30: Sure. Please propose upstream.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -35,10 +39,10 @@ MAINPACKAGE=./cmd/manager
 # Containers may default GOFLAGS=-mod=vendor which would break us since
 # we're using modules.
 unexport GOFLAGS
-# rporresm - don't set this to Linux, there are Mac OS X in the world
+# upstream-diff - don't set this to Linux, there are Mac OS X in the world
```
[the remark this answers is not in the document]
rporres: I didn't propose this to upstream as I had to hack a bit my way around the docker build target... We'll take a further look to see if I can come up with something nice

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -35,10 +39,10 @@ MAINPACKAGE=./cmd/manager
...
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
-# rporresm - we want to use vendor
+# upstream-diff - we want to use vendor
```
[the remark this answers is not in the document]
rporres: I agree

boilerplate/openshift/golang-osd-operator/standard.mk
rporres: Yeah, I had to do that but I found it awful enough to avoid suggesting it in upstream 😂  But if you say so, we will

2uasimojo: Yeah, as long as we have some good comments around it, including acknowledging that it's temporary until we figure out how to containerize local builds, it'll be fine.

boilerplate/openshift/golang-osd-operator/standard.mk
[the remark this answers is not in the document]
2uasimojo: I think I get what we're trying to do: when we're validating, we want the env appropriate to where we're running; but when we're actually building, we always want to build for linux. Again, this is something that will be moot once we figure out containerized builds, but I wonder if in the meantime it's worth having separate `GOENV_LOCAL` and a `GOENV_BUILD` vars. The latter would only be used [below](https://github.com/app-sre/deployment-validation-operator/pull/37/files/f365d7c1acc58822e358892fadea577b5e81011f#diff-6efb8dd653c9062efcfb26adab7756fcR106). WDYT?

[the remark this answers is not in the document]
2uasimojo: Or actually, what if we explicitly set `GOOS=linux` in `GOBUILDFLAGS` (with appropriate comment)?

2uasimojo:
At this point, suggest [updating boilerplate](https://github.com/app-sre/deployment-validation-operator/pull/38) and then rebasing this PR.

rrati:
I'd rather do the boilerplate update in a separate PR.  This is shifting enough around I'd rather isolate problems (if any) to just this change.

2uasimojo:
> I'd rather do the boilerplate update in a separate PR. This is shifting enough around I'd rather isolate problems (if any) to just this change.

That's what I meant. See the linked PR (#38).

rporres:
We will not follow this path. See https://github.com/app-sre/deployment-validation-operator/pull/41 instead

PR #38, opened 2020-10-12
Boilerplate: Update to f435f4ab6213d69d6ffbe493435e8b062cf54684

2uasimojo:
Conventions:
- openshift/golang-osd-operator: Update
---
https://github.com/openshift/boilerplate/compare/ebfd04e49747614eb6cb0c046c338f246a9df35d...f435f4ab6213d69d6ffbe493435e8b062cf54684

commit: aafa039ecc2bb506c38b29fd3bbd7f31819bf52c
author: Rob Rati
Updates to standard.mk to support building on other platforms and broke down operator IMG into usable sub-components

commit: ee247ba8d7b1b92fa17776d027847aeb819008ea
author: Eric Fried
Use 7 characters for the commit SHA

...to align with what appsre is using.

commit: 93b5e68b8695162780ea77226e62f4e827a2c992
author: Eric Fried
Add in-repo prow `build_root` configuration

commit: f92583fc3a06e517f9aaf08e57f6b94aa4c67744
author: Eric Fried
Move `generate-check` under `validate`

...as was originally intended (see the referenced Jira).

Jira: [OSD-4673](https://issues.redhat.com/browse/OSD-4673)

commit: 3b93083cf02aecf21db7420f220e5f3a6168a7dc
author: Eric Fried
Make the default target work

Some versions of `make` don't honor the `default` target. Others don't
pay attention to `.DEFAULT_GOAL`. Fix standard.mk in the
openshift/golang-osd-operator convention to accommodate all of them.

commit: 043bae4f0d6907fe3fb1c8457a4e62b069861f06
author: Eric Fried
Make `make validate` include `freeze-check`

...to ensure boilerplate-owned files are not changed.

This adds a `make` target in the boilerplate framework itself called
`boilerplate-freeze-check` and includes it in the chain of the
`validate` target in the `openshift/golang-osd-operator` convention.

Jira: [OSD-5366](https://issues.redhat.com/browse/OSD-5366)

commit: 0c40e4de3f877457704b58fc53cf72faddfd5aaf
author: Eric Fried
Review updates from PR #56

- Soften README wording around using `make boilerplate-commit` to make
it sound less mandatory.
- Remove `go.mod`, `go.sum`, and per-convention
`.boilerplate-ignore-files` from the ignore list when ensuring a "clean"
checkout prior to `update`. These were causing a test failure, but the
right thing was to fix the test, which this commit does.
- Add some comments around non-obvious chunks of code.
- Factor out `BOILERPLATE_GIT_REPO` and `BOILERPLATE_GIT_CLONE`
defaulting. (Note that this still isn't used by the main `update`
driver, because that needs to be self-contained for bootstrapping.)
- Use common.sh for freeze-check, as part of the above.

commit: 28da926566d421d5973b8579e7cf6b21b3df99d2
author: Naveen Malik
Don't break lint if there is no yaml to validate.

Make targets pass in files to validate but not all repos have yaml that needs validating.
In that case the yaml validation fails because it prints usage and exits with 1.

This PR simply changes the exit code to 0 so it's not going to fail.

There is no great alternative to this, we don't know where yaml might be or if there might be a directory to pass in etc.  Proposing we simply don't fail without arguments.

commit: 7c8928c70ec5dd216516cbe3d2c59a750f0eef57
author: Arjun Naik
Added a make target to build and push the boilerplate image

Signed-off-by: Arjun Naik <anaik@redhat.com>

commit: b24e8d5d094a339ad622589e42f3790568e54130
author: Benjamin Dematteo
Integrating remaining comments from PR#41 review

commit: 9c5adcc6a9026422324e77d0fbf8c4c3dc2df90f
author: Eric Fried
Script/target to generate a commit after update

This adds a `boilerplate-commit` target brought in by the nexus
makefile, intended for use by consuming repositories. The target invokes
a script of the same name that creates a git commit out of your current
unstaged changes if they relate to boilerplate bootstrap, update, and/or
subscriptions. When the changes include an update, the generated commit
message contains a log summary of the boilerplate commits brought in by
the update, as well as a github `compare` URL for the same, in the
spirit of promote.sh.

commit: 039f40dea15739c856a5568707a9d14912bb9cd6
author: Eric Fried
Configure golangci-lint to keep going

By default, golangci-lint will stop printing errors after a certain
limit. This can be annoying, especially in CI, as you would have to keep
running it repeatedly to fix all the problems. So this commit adds
configuration values to disable those limits, meaning *all* errors will
be printed on each run.

commit: 7c3a564e8524fc827dc00e0fddcabaf8f12ba5bb
author: Eric Fried
Disallow (most) dirty checkouts when updating

Since boilerplate update messes with your local checkout, it can be hard
to recover pre-existing local changes if something goes wrong and you
need to back out. With this commit, update fails quickly if your
checkout is not clean.

However, to facilitate the most common workflows, we allow the following
exceptions:
- We try to detect when you're bootstrapping, and allow only the files
related to that.
- We try to detect when you're only updating the config, i.e. changing
subscriptions, and allow that.

commit: 237207a4900095092aa31ec9b0de1edabc084bdf
author: Eric Fried
Refactor: Rename local vars in update

The variable names in the main update script were ambiguous and confused
me every time I had to deal with the script. Change them to be more
developer-friendly.

commit: b41514107d7632063be4afd2279fdb5dcffa9670
author: Benjamin Dematteo
Adding a check_generation make target to check generated files are not altered

commit: 601c37ed1ce04801cce664f4b97052fe993c7ebe
author: Arjun Naik
Fixed where git is installed in image

Signed-off-by: Arjun Naik <anaik@redhat.com>


boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -74,17 +78,17 @@ docker-build: isclean
...
 
 .PHONY: push
 push: docker-push
 
 .PHONY: go-check
-go-check: # golanci-lint and other static analysis
+go-check: ## Golang linting and other static analysis
 	${CONVENTION_DIR}/ensure.sh golangci-lint
-	GOLANGCI_LINT_CACHE=${GOLANGCI_LINT_CACHE} golangci-lint run -c ${GOLANGCI_LINT_CONFIG} ./...
+	GOLANGCI_LINT_CACHE=${GOLANGCI_LINT_CACHE} golangci-lint run -c ${CONVENTION_DIR}/golangci.yml ./...
```
rporres, line 91: This is going to override our own config that it was possible to set before... I'll fix it in a subsequent PR

rporres (approved):
LGTM. Thanks for doing this, @2uasimojo

boilerplate/openshift/golang-osd-operator/standard.mk
2uasimojo: We're trying to avoid allowing too much customization by consumers, as it sort of defeats the purpose. What did you have in mind?

rporres: I would not assume every consumer of boilerplate will have the same ideas about linting and static code analysis, e.g

* https://github.com/openshift/aws-account-operator/blob/master/.golangci.yml
* https://github.com/openshift/hive/blob/master/golangci.yml
* https://github.com/openshift/telemeter/blob/master/.golangci.yml
* https://github.com/app-sre/deployment-validation-operator/blob/master/.golangci.yml

2uasimojo: Right now they don't, I agree. Converging on that is one of the primary pushes of the boilerplate effort. Common linting rules across repositories has $benefits :)

rporres: This is what I had in mind for boilerplate: https://github.com/app-sre/deployment-validation-operator/pull/39

PR #39, opened 2020-10-14
Allow custom linter config

rporres:


2uasimojo:
Ah, so bp isn't going to let you do this :)

Since you last looked, we added a check to prevent consumers from editing bp-controlled files.

rporres:
`make build` and `make docker-build` work for me locally 😅

2uasimojo:
Part of the package will be standardizing prow configs, one of which will be to run `make validate`, which runs `freeze-check`, which will fail for the reason cited above.

This would be an eminently sane thing to propose to boilerplate itself... if we were willing to allow customizations like this. Perhaps you can do that and we can have the discussion there?

rporres:
Done in https://github.com/openshift/boilerplate/pull/74

2uasimojo:
Responded in https://github.com/openshift/boilerplate/pull/74.

I'm curious why you want to do this, since you're already lint-free using the (stricter) config in the convention. YAGNI.

rporres:
We won't follow this path. See https://github.com/openshift/boilerplate/pull/74 to get more details

PR #47, opened 2020-10-28
Check that bundle and catalog are well built

rporres:
* Bundle is checked using opm.
* Catalog needs to run so that we can query it. Let's see how this
  goes, docker from a script is a bit fragile.


rrati (approved):
lgtm

2uasimojo:
This is cool.

FYI I'm working on making sure you have the desired versions of opm and grpcurl in the boilerplate backing image, so you'll be able to remove chunks like [this one](https://github.com/app-sre/deployment-validation-operator/pull/47/files#diff-e4d1332c8c88515a1a754d45afa9784b39bd05390c423a13dc6ca42985d99198R190-R207).

PR #52, opened 2020-10-30
Run additional golangci checks

rporres:
These will be run on top of what boilerplate offers. See https://github.com/openshift/boilerplate/pull/92

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,22 +4,12 @@ run:
 linters:
   disable-all: true
   enable:
-  - deadcode
-  - errcheck
-  - gosimple
-  - govet
-  - ineffassign
-  - staticcheck
-  - structcheck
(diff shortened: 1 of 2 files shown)


rporres:
cc @2uasimojo @rrati

2uasimojo (commented):
Yeah, this looks legit to me.

Why are you removing all those other checks?

rporres:
> Why are you removing all those other checks?

Because they are already covered by the standard boilerplate config

rrati (approved):
lgtm

2uasimojo:
> > Why are you removing all those other checks?
> 
> Because they are already covered by the standard boilerplate config

Ah, okay, I guess that makes sense... unless we decide to remove them from bp. Which I doubt we'll do, so cool.

PR #56, opened 2020-11-06
Proposal single channel

jmelis:
Yesterday we met with the OLM team and the introduced us the `semver` approach. It seems good at first sight for us, however there are two things that we can't fix with the `semver` approach:

- The appeal of using `semver` is that you don't need to store state, you only need to fetch the latest index image and add a new bundle to it. However, we still need state in order to be able to rebuild the index image in a disaster event in the container registry.
- If we use `semver` we can't easily skip a faulty version. Let's say we find a bundle to be faulty. In this scenario we want to ensure that the next time we build a bundle it jumps over the fault version. The `skips` parameter is ignored in the `semver` mode. Maybe we could use `skipRange` (which is respected at runtime in `semver` mode), but we haven't explored this.

Given these two reasons we propose to continue using the `replaces` approach, with two changes:

1. We stop pruning versions, because the ultimate goal is to **run the same payload in stage and in prod**. The consequence of this is that we only need one channel now, therefore only one branch (this is reflected in the PR, see commit https://github.com/jmelis/deployment-validation-operator/commit/6138d2cb1c5a1e3d4513dfa6d7494d2cd3a05a2f).
1. Provide a process to skip a faulty bundle. This is the process:
    - Go to the saas bundle repo and append `SKIP` at the end of the `bundle_versions_file`. Next time the CI job is run, once the code is fixed, it will skip that version and make the new bundle replace the last healthy one.
    - In the staging cluster manually delete the csv, catalogsource and subscription objects. That way when the fixed catalog lands it will install the latest version.

The whole point of this proposal is that we are using the same payload for staging and production, which means we are testing the upgrade path. Note that it also simplifies the build process a bit because we are removing the bundle pruning logic.

### DIFF
--- a/deploy/bundle/generate-csv-template.py
+++ b/deploy/bundle/generate-csv-template.py
@@ -5,6 +5,13 @@ import os
 import sys
 import yaml
 import pathlib
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-r", "--replaces", help="Replaces version", type=str)
+parser.add_argument('-s','--skip', action='append',
+        help='Skips version (can be specified multiple times)')
+args = parser.parse_args()
 
 root = pathlib.Path(__file__).parent.absolute() / '../..'
 manifest_dir = root / 'deploy/openshift'
@@ -46,6 +53,15 @@ with open(manifest_dir / 'operator.yaml', 'r') as stream:
 
 csv['spec']['install']['spec']['deployments'][0]['spec']['template']['spec']['containers'][0]['image'] = \
     '${IMAGE}:${IMAGE_TAG}'
+
+if args.replaces:
+    csv['spec']['replaces'] = f'deployment-validation-operator.v{args.replaces}'
+
+if args.skip:
+    csv['spec']['skips'] = [
+        f'deployment-validation-operator.v{version}' for version in
+        sorted(args.skip, key=lambda v: int(v.split('.')[2].split('-')[0]))
+    ]
 
 now = datetime.datetime.now()
 csv['metadata']['annotations']['createdAt'] = \
--- a/hack/build_channel_catalog.sh
+++ b/hack/build_channel_catalog.sh
@@ -1,7 +1,19 @@
 #!/bin/bash
 
+set -euo pipefail
+
+# General vars to modify behaviour of the script
+DRY_RUN=${DRY_RUN:-false}
+DELETE_TEMP_DIR=${DELETE_TEMP_DIR:-true}
+BRANCH=${BRANCH:-master}
+SET_X=${SET_X:-false}
+
+[[ "$SET_X" != "false" ]] && set -x
+
 function log() {
-    echo "$(date "+%Y-%m-%d %H:%M:%S") -- $1"
+    msg="$(date "+%Y-%m-%d %H:%M:%S")"
+    [[ "$DRY_RUN" != "false" ]] && msg="$msg -- DRY-RUN"
+    echo "$msg -- $1"
 }
 
 count=0
@@ -17,13 +29,9 @@ for var in BUNDLE_IMAGE \
            GOOS \
            GOARCH \
            OPM_VERSION \
-           BUNDLE_VERSIONS_REPO \
-           BRANCH_CHANNEL \
-           APP_INTERFACE_USERNAME \
-           APP_INTERFACE_PASSWORD \
-           APP_INTERFACE_BASE_URL
+           BUNDLE_VERSIONS_REPO
 do
-    if [ ! "${!var}" ]; then
+    if [ ! "${!var:-}" ]; then
       log "$var is not set"
       count=$((count + 1))
     fi
@@ -31,18 +39,17 @@ done
 
 [[ $count -gt 0 ]] && exit 1
 
-# We shouln't need set -u as we have checked for the vars before
-# but let's play safe
-set -euo pipefail
+# Check we are running an opm supported container engine
+image_builder=$(basename "$CONTAINER_ENGINE")
+if [[ "$image_builder" != "docker" && "$image_builder" != "podman" ]]; then
+    # opm error messages are obscure. Let's make this clear
+    log "image_builder $image_builder is not one of docker or podman"
+    exit 1
+fi
 
-log "Building $OPERATOR_NAME version $OPERATOR_VERSION in branch $BRANCH_CHANNEL"
+log "Building $OPERATOR_NAME version $OPERATOR_VERSION"
 
-# General vars to modify behaviour of the script
-DRY_RUN=${DRY_RUN:-false}
-DELETE_TEMP_DIR=${DELETE_TEMP_DIR:-true}
-REMOVE_UNDEPLOYED=${REMOVE_UNDEPLOYED:-false}
-
-temp_dir=$(mktemp -d --suffix "-$(basename $0)")
+temp_dir=$(mktemp -d --suffix "-$(basename "$0")")
 [[ "$DELETE_TEMP_DIR" == "true" ]] && trap 'rm -rf $temp_dir' EXIT
 
 engine_cmd="$CONTAINER_ENGINE --config=$CONFIG_DIR"
@@ -58,52 +65,46 @@ else
     bundle_versions_repo_url="https://$BUNDLE_VERSIONS_REPO"
 fi
 
-log "Cloning $BRANCH_CHANNEL branch from $BUNDLE_VERSIONS_REPO into $saas_operator_dir_base"
-git clone --branch "$BRANCH_CHANNEL" "$bundle_versions_repo_url" "$saas_operator_dir_base"
+log "Cloning $BUNDLE_VERSIONS_REPO into $saas_operator_dir_base"
+git clone --branch "$BRANCH" "$bundle_versions_repo_url" "$saas_operator_dir_base"
 
+# if the line contains SKIP it will not be included
 prev_operator_version=""
-removed_versions=""
+prev_good_operator_version=""
+skip_versions=()
 if [[ -s "$bundle_versions_file" ]]; then
     log "$bundle_versions_file exists. We'll use to determine current version"
-    if [[ "$REMOVE_UNDEPLOYED" == "true" ]]; then
-        log "Checking if we have to remove any versions more recent than deployed hash"
-        # this will need to be modified if at any point we move to a canary like deployment
-        # where not all environments are deployed with the same operator version
-        deployed_hash=$(
-            curl -s -H "Authorization: Basic $(echo -n $APP_INTERFACE_USERNAME:$APP_INTERFACE_PASSWORD | base64)" \
-                -g "https://$APP_INTERFACE_BASE_URL/graphql?query={saas_files:saas_files_v1{name,resourceTemplates{name,targets{namespace{environment{name,labels}},ref}}}}" | \
-                jq -r '.data.saas_files[] | select(.name=="saas-'$OPERATOR_NAME'") | .resourceTemplates[].targets[] | select(.namespace.environment.labels | contains("\"type\":\"production\"")) | .ref' | \
-                uniq
-        )
 
-        log "Current deployed hash is $deployed_hash"
+    prev_operator_version=$(tail -n 1 "$bundle_versions_file" | awk '{print $1}')
 
-        new_bundle_versions_file=$(mktemp -p "$temp_dir" new_bundle_versions_file.XXXX)
-        delete=false
-        # Sort based on commit number
-        for version in $(sort -t . -k 3 -g "$bundle_versions_file"); do
-            if [[ "$delete" == false ]]; then
-                echo "$version" >> "$new_bundle_versions_file"
+    # we traverse the bundle versions file backwards
+    # we cannot use pipes here or we would lose the inner variables changes
+    while read -r line; do
+        if [[ "$line" == *SKIP* ]]; then
+            version=$(echo "$line" | awk '{print $1}')
+            skip_versions+=("$version")
+        else
+            prev_good_operator_version="$line"
+            break
+        fi
+    done < <(sort -r -t . -k 3 -g "$bundle_versions_file")
 
-                short_hash=$(echo "$version" | cut -d- -f2)
-
-                if [[ "$deployed_hash" == "${short_hash}"* ]]; then
-                    log "found deployed hash in the bundle versions repository"
-                    delete=true
-                fi
-            else
-                log "Adding $version to removed versions"
-                removed_versions="$version $removed_versions"
-            fi
-        done
-        [[ -n "$removed_versions" ]] && log "The following versions will be removed: $removed_versions"
-        cp "$new_bundle_versions_file" "$bundle_versions_file"
+    if [[ -z "$prev_good_operator_version" ]]; then
+        # This means that we have skipped all the available versions. In this case we're going to use the last
+        # SKIP version as the prev_good_operator_version to have something to feed replaces in the CSV
+        log "No unskipped version in $bundle_versions_file. We'll use the last skipped one: ${skip_versions[0]}"
+        prev_good_operator_version="${skip_versions[0]}"
     fi
-    prev_operator_version=$(tail -n 1 "$bundle_versions_file")
+
     log "Previous operator version is $prev_operator_version"
+    log "Previous good operator version is $prev_good_operator_version"
+    [[ ${#skip_versions[@]} -gt 0 ]] && log "We will be skipping: ${skip_versions[*]}"
 else
     log "No $bundle_versions_file exist. This is the first time the operator is built"
-    mkdir -p "$(dirname $bundle_versions_file)"
+    if [[ ! -d "$(dirname "$bundle_versions_file")" ]]; then
+        log "Operator directory doesn't exist in versions repository. Exiting"
+        exit 1
+    fi
 fi
 
 if [[ "$OPERATOR_VERSION" == "$prev_operator_version" ]]; then
@@ -112,26 +113,28 @@ if [[ "$OPERATOR_VERSION" == "$prev_operator_version" ]]; then
 fi
 
 # image:tag definitions
-bundle_image_current_commit="$BUNDLE_IMAGE:${BRANCH_CHANNEL}-$CURRENT_COMMIT"
-bundle_image_latest="$BUNDLE_IMAGE:${BRANCH_CHANNEL}-latest"
-catalog_image_current_commit="$CATALOG_IMAGE:${BRANCH_CHANNEL}-$CURRENT_COMMIT"
-catalog_image_latest="$CATALOG_IMAGE:${BRANCH_CHANNEL}-latest"
+bundle_image_current_commit="$BUNDLE_IMAGE:$CURRENT_COMMIT"
+bundle_image_latest="$BUNDLE_IMAGE:latest"
+catalog_image_current_commit="$CATALOG_IMAGE:$CURRENT_COMMIT"
+catalog_image_latest="$CATALOG_IMAGE:latest"
 
 # Build bundle
 mkdir -p "$MANIFEST_DIR"
 csv_template=$(mktemp -p "$temp_dir" csv_template.XXXX)
-./"$BUNDLE_DEPLOY_DIR"/generate-csv-template.py > "$csv_template"
+generate_csv_template_args=""
+[[ -n "$prev_good_operator_version" ]] && generate_csv_template_args="--replaces $prev_good_operator_version"
+if [[ ${#skip_versions[@]} -gt 0 ]]; then
+    for version in "${skip_versions[@]}"; do
+        generate_csv_template_args="$generate_csv_template_args --skip $version"
+    done
+fi
+
+./"$BUNDLE_DEPLOY_DIR"/generate-csv-template.py $generate_csv_template_args > "$csv_template"
 oc process --local -o yaml --raw=true \
     IMAGE="$OPERATOR_IMAGE" \
     IMAGE_TAG="$OPERATOR_IMAGE_TAG" \
     VERSION="$OPERATOR_VERSION" \
-    REPLACE_VERSION="$prev_operator_version" \
     -f "$csv_template" > "$CSV"
-
-if [[ "$prev_operator_version" == "" ]]; then \
-    sed -i.bak "/ *replaces:/d" "$CSV"
-    rm -f "$CSV.bak"
-fi
 
 # We need an up-to-date version of opm executable
 opm_local_executable=$(which opm || true)
@@ -147,13 +150,6 @@ else
     curl -s -L "$opm_download_url" -o "$temp_dir/opm"
     chmod u+x "$temp_dir/opm"
     opm_local_executable="$temp_dir/opm"
-fi
-
-image_builder=$(basename "$CONTAINER_ENGINE")
-if [[ "$image_builder" != "docker" && "$image_builder" != "podman" ]]; then
-    # opm error messages are obscure. Let's make this clear
-    log "image_builder $image_builder is not one of docker or podman"
-    exit 1
 fi
 
 # opm won't get anything locally, so we need to push the bundle even in dry run mode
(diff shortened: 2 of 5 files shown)


rporres:
So this replaces the automatic pruning which get us with untested upgrade paths by the manual intervention with the SKIP keyword. It definitely makes sense, has benefits and I can see it working. I guess we would need input from @2uasimojo and @jharrington22 to move forward with this approach.

As for the PR per se, we would need to update the https://github.com/app-sre/deployment-validation-operator/blob/master/deploy/openshift/deployment-validation-operator-olm.yaml to remove the `CHANNEL` usage.

hack/build_channel_catalog.sh
```diff
@@ -58,48 +64,13 @@ else
     bundle_versions_repo_url="https://$BUNDLE_VERSIONS_REPO"
 fi
 
-log "Cloning $BRANCH_CHANNEL branch from $BUNDLE_VERSIONS_REPO into $saas_operator_dir_base"
-git clone --branch "$BRANCH_CHANNEL" "$bundle_versions_repo_url" "$saas_operator_dir_base"
+log "Cloning $BUNDLE_VERSIONS_REPO into $saas_operator_dir_base"
+git clone "$bundle_versions_repo_url" "$saas_operator_dir_base"
```
2uasimojo, line 68: So this is going to start using `master`, which we currently ignore.

- Something normally creates the `staging` and `production` branches. I'm not sure if it's part of the app-interface automation triggered by [this config](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/osd-operators/app.yml) or if it's done by (the first run of) app_sre_build_deploy.sh for a given operator. We should look into that. (Not that having those branches hurts anything; but it's noise.)
- This repository is bootstrapping, so it's a non-issue; but as we look to using this as our example for established projects, we'll need to come up with some kind of transition plan.

hack/build_channel_catalog.sh
```diff
@@ -58,48 +64,13 @@ else
...
-            else
-                log "Adding $version to removed versions"
-                removed_versions="$version $removed_versions"
-            fi
-        done
-        [[ -n "$removed_versions" ]] && log "The following versions will be removed: $removed_versions"
-        cp "$new_bundle_versions_file" "$bundle_versions_file"
-    fi
-    prev_operator_version=$(tail -n 1 "$bundle_versions_file")
+    prev_operator_version=$(grep -v SKIP "$bundle_versions_file" | tail -n 1)
```
2uasimojo, line 73: From what I can tell, this will only allow us to `SKIP` one or more of the *latest* versions. In other words, editing my bundle versions file from

```
v1.0.0
v1.0.1
v1.0.2
v1.0.3
```
to
```
v1.0.0
v1.0.1
v1.0.2 SKIP
v1.0.3 SKIP
```
will cause my build of `v1.0.4` to use `replaces: v1.0.1`. Cool.

But if I do something like this:
```
v1.0.0
v1.0.1
v1.0.2 SKIP
v1.0.3
```
it'll be a no-op. Specifically, it will *not* cause the index to be rebuilt such that `v1.0.3` replaces `v1.0.1`.

I recognize that this would be a tough one to solve; and maybe we don't want to do it right away. But we should at least leave a detailed explanation in the README and/or in a comment here.

2uasimojo (commented):
Yeah, this seems legit to me.

I love how much it simplifies the logic.

Left some commentary inline. It's not necessarily anything that should prevent us merging this PR though.

hack/build_channel_catalog.sh
rporres: Some saas-*-bundle repositories have been created in this file: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/dependencies/gitlab/gitlab.yml. I know nothing about any automation creating branches and directories in those repositories, I don't think there's any. [app_sre_create_image_catalog.sh](https://github.com/openshift/aws-account-operator/blob/master/hack/app_sre_create_image_catalog.sh) assumes both the branch and the directory exist in the saas repo.

I'm fine with staying in `master`. I'm not fine with continuing if there's no specific directory for the operator since the versions repo is going to be shared and we need to differentiate the case of a first run from targetting the wrong versions repository.

hack/build_channel_catalog.sh
rporres: So you want to skip a version after a newer one has already been deployed... That's not going to work, indeed... The current system will work as expected if we SKIP every version after the wrong one

rporres:
There's something in OLM that I don't get. The documentation asks to use both `replaces` and `skip` when you want to skip a version: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/how-to-update-operators.md#skipping-updates. In the current proposal we don't add the `skip` field.

hack/build_channel_catalog.sh
rrati: Actually, I don't think it'll matter except in the case of trying to rebuild a catalog from scratch. In the above example, bundle 1.0.3 will be built with replaces 1.0.1 because when the 1.0.3 bundle was created the history looked like:

```shell
v1.0.0
v1.0.1
v1.0.2 SKIP
```

When bundle 1.0.4 is created, the history will look like this:

```shell
v1.0.0
v1.0.1
v1.0.2 SKIP
v1.0.3
```

So the last version will be selected at v1.0.3 for the 1.0.4 bundle, which is correct.  The 1.0.3 bundle is already in the catalog with the correct replaces (v1.0.1) and we aren't recreating that bundle.

If we really need this exact history though, I think this can be accomplished by changing the logic slightly.  Create a new set of versions from the existing one, taking out each entry that has SKIP next to it.  It should basically be a list of versions created like:

grep -v SKIP "$bundle_versions_file" 

That would get us a valid upgrade chain accepting any number of SKIP versions.  However, since we are only ever creating the latest bundle except in an emergency I'm not sure where this logic would even be used in this script.

rrati:
> There's something in OLM that I don't get. The documentation asks to use both `replaces` and `skip` when you want to skip a version: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/how-to-update-operators.md#skipping-updates. In the current proposal we don't add the `skip` field. Unless there's a good reason to not follow the doc, I guess we should look into adding the `skip` field

In our testing if you have a catalog like this:
v1.0.0
v1.0.1 replaces v1.0.0
v1.0.2 replaces v1.0.0

Then OLM will skip from v1.0.0->v1.0.2 without starting the operator at v1.0.1.  I don't know how adding the skip field changes the internal logic, but if we can successfully add the skip then I don't think it's a bad thing to add.

rporres:
>  I don't know how adding the skip field changes the internal logic, but if we can successfully add the skip then I don't think it's a bad thing to add.

Me neither. It's just that it is there in the example that is exactly as the one you're describing 😅

hack/build_channel_catalog.sh
rrati: Rewriting the skips list in a CSV is basically rewriting history.  I don't see how it can be done with automation.  The bundles have to be regenerated with the same bundle version/sha combination to avoid breaking the upgrade change.  This could be done manually of course, but then the git repo would not reflect what is actually in a bundle.  To have the repo reflect the corrected bundle images you would probably have to re-write git history for those versions we need to skip.  Then you have to address the old bundle image already existing on the openshift nodes or else the corrected bundle images won't be pulled.  If that isn't done as well then the upgrade chain is not going to be tested in stage as it would be implemented in prod.  IMO, this is worse than the heavy-handed approach @rporres suggested.

I'm not convinced this issue is a major concern we need to plan to address.  IMO, @rporres has a means to handle it should it occur.  The recovery process ensures that the upgrade chain will be tested in stage before being promoted to pod.

2uasimojo: > Rewriting the skips list in a CSV is basically rewriting history. I don't see how it can be done with automation.

I don't think this is right. We are dropping a fresh CSV every time we run this script. That's the thing we're computing `replaces` for. What we're talking about here is _also_ computing `skips` for the same CSV. And writing a different `skips` list in the current CSV while leaving the old CSVs alone is exactly the use case described in the doc -- unless I'm wildly misunderstanding it.

rrati: Honestly the doc isn't really clear (to me) for the use case we are trying to solve.  If you look at the example, X replaces X-2 and skips X-1.  To me, that meant we can't have a valid chain of X, X+1, X+2, X+3 (skip X+1) because the replaces in X+2 will be wrong as it would need to replace X, not X+1.

However, talking with the OLM team my understanding appears to be incorrect and yours appears to be correct.

One thing to be cautious about with skips, is that it will be ignored if we ever move (or are forced to move to) semantic versioning.  We might want to use skipRange instead as it would still be honored if we ever move to semantic versioning.

rrati: Update on this.  After further discussion with the OLM team, it seems my interpretation for skips is correct.  It can only be used to skip n-1 and can not be used to define a list of previous versions to skip.  This is the same for skipRange.  There currently isn't a way to have the latest catalog define a list of previous versions to skip.

2uasimojo: I concur, with one quibble:

> n-1

It can be any number of contiguous versions from n-1 back. (If it's just n-1, it's technically unnecessary, redundant with `replaces` pointing at n-2.)

So we should set `skipRange` to cover the skipped version(s), as that'll get us happy with opm+semver. We could just always set it to all the known skips, because that's super easy to compute, and doesn't do any harm even when some values would be ignored. But at a minimum we should set it to cover the contiguous range up to n-1 iff n-1 is itself a skip.

A handy detail: `skipRange` accepts semver range semantics, so we could set it as (pseudocode) `skipRange = join(' || ', skiplist)`. That is, we don't need to compute inequalities; we can just enumerate the specific versions with logical OR.

rporres: I've never understood the documentation of `skips` as all the versions to skip but to skip versions between two certain versions... a list instead of a `skipRange`. Exactly as the last message from @rrati

rrati: @2uasimojo I don't think that is correct.  We can only set skips for csv after the one being skipped.  It's not possible to have the latest csv define a list of versions to skip.  The latest CSV can not define n-2 and n-1 skips if n-2 csv is already in the bundle and n-1 does not skip it.  Per the conversion:

We want the latest bundle in a catalog to determine which previous versions need to be skipped.  That way when we install the catalog fresh (as part of recovery) it will start at X-n and upgrade through all the operator versions except the ones listed in the skip of the latest bundle.

Yeah it's not possible today. We have some designs about how altering this package level metadata could work, but nothing is implemented yet.

Am I misunderstanding you?

rporres: If I have understood https://coreos.slack.com/archives/C3VS0LV41/p1605117420420500 correctly, the only way to currently address the scenario in which we push something after the one we have to skip is the workaround I proposed. Is correct @rrati @2uasimojo?

rporres:
I've added the CSV skips logic in d071e17 to the best of my understanding of yesterday's discussion. Before merging this we would need to:

1. Create a new tag of the last catalog to remove the `staging-` or `production-` bit
2. Create the directory `deployment-validation-operator` in `master`

hack/build_channel_catalog.sh
```diff
@@ -17,32 +26,27 @@ for var in BUNDLE_IMAGE \
...
+fi
 
-# General vars to modify behaviour of the script
-DRY_RUN=${DRY_RUN:-false}
-DELETE_TEMP_DIR=${DELETE_TEMP_DIR:-true}
-REMOVE_UNDEPLOYED=${REMOVE_UNDEPLOYED:-false}
+log "Building $OPERATOR_NAME version $OPERATOR_VERSION"
 
-temp_dir=$(mktemp -d --suffix "-$(basename $0)")
+temp_dir=$(mktemp -d --suffix "-$(basename "$0")")
```
2uasimojo, line 49: Curious why this delta. The result will almost always be the same. In particular, you'll still break if there are spaces in `$0` (but why would there be?). Did you mean
```suggestion
temp_dir=$(mktemp -d --suffix "-$(basename '"$0"')")
```
?

hack/build_channel_catalog.sh
```diff
@@ -58,52 +62,43 @@ else
...
+        if [[ $(echo "$line" | grep SKIP) == "" ]]; then
+            prev_good_operator_version="$line"
+            break
+        else
+            version=$(echo "$line" | awk '{print $1}')
+            skip_versions+=("$version")
+        fi
+    done < <(sort -r -t . -k 3 -g "$bundle_versions_file")
+
+    if [[ -z "$prev_good_operator_version" ]]; then
```
2uasimojo, line 88: Oh, obviates the corner case I mentioned below. Not sure it's legit to bail here, though. If the versions file contains only

```
v1.0.0 SKIP
```
...don't we want to use `skipRange` but no `replaces`?

We should probably ask the OLM folks on this.

deploy/bundle/generate-csv-template.py
```diff
@@ -47,6 +54,15 @@
 csv['spec']['install']['spec']['deployments'][0]['spec']['template']['spec']['containers'][0]['image'] = \
     '${IMAGE}:${IMAGE_TAG}'
 
+if args.replaces:
+    csv['spec']['replaces'] = f'deployment-validation-operator.v{args.replaces}'
+
+if args.skip:
+    csv['spec']['skips'] = [
```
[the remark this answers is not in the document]
rporres: Does `skipRange` work with `replaces`? It was my understanding it didn't or at least it looks to me as a weird mixture. Since we want to continue using `replaces` I prefer to use `skip`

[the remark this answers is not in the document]
rrati: iirc, semver mode ignores `replaces` and relies on the ordering based upon the operator's version field.  It's a bit deceptive.  Semver mode uses the operator.version field to determine updates.  Non-semver mode use operator.name and replaces (which keys off operator.name) to determine upgrade chain.  They are similar, but different in important ways.

hack/build_channel_catalog.sh
```diff
@@ -65,16 +63,42 @@ else
...
+    fi
+
     log "Previous operator version is $prev_operator_version"
+    log "Previous good operator version is $prev_good_operator_version"
+    [[ ${#skip_versions[@]} -gt 0 ]] && log "We will be skipping: ${skip_versions[*]}"
 else
     log "No $bundle_versions_file exist. This is the first time the operator is built"
-    mkdir -p "$(dirname $bundle_versions_file)"
+    if [[ ! -d "$(dirname "$bundle_versions_file")" ]]; then
+        log "Operator directory doesn't exist in versions repository. Exiting"
```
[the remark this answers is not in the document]
rporres: Mostly to ensure that we're not writing in the repo that we didn't want. It is very critical to make sure that whenever we've written and push we're ok.

hack/build_channel_catalog.sh
```diff
@@ -91,19 +115,21 @@ catalog_image_latest="$CATALOG_IMAGE:latest"
 # Build bundle
 mkdir -p "$MANIFEST_DIR"
 csv_template=$(mktemp -p "$temp_dir" csv_template.XXXX)
-./"$BUNDLE_DEPLOY_DIR"/generate-csv-template.py > "$csv_template"
+generate_csv_template_args=""
+[[ -n "$prev_operator_version" ]] && generate_csv_template_args="--replaces $prev_good_operator_version"
```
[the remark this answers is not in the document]
rporres: 100% right. It was a bug

hack/build_channel_catalog.sh
rporres: I was making `shellcheck` happy by quoting every single variable that has to be quoted

hack/build_channel_catalog.sh
rporres: you're indeed correct, but If you want to skip your only version... you may want to start from scratch.

The idea was to continue using `replaces`, I thought there was an agreement on that since it's one of the foundations of this PR.

rrati: If the versions file only contains a SKIP entry then we will be creating a new catalog from scratch.  There shouldn't be a replaces used at all in that case.

2uasimojo: Summary of that discussion:
Specifically in appsre it doesn't matter because the process of purging bad versions is manual.
but
In general, setting `replaces` to the latest skipped version -- the only time `replaces` and `skipRange` overlap -- would provide a path from the bad version if it's already installed somewhere, but also ensure that that version isn't available for fresh installation.
so
IMO we should do the latter anyway, since it's harmless for appsre but more generally correct.

hack/build_channel_catalog.sh
2uasimojo: except what you've done here is *un*quote `$0`.

rporres: bash doesn't agree with you :)

```
[rporresm@rporresm-mac test]$ cat example\ with\ spaces.sh
temp_dir=$(mktemp -d --suffix "-$(basename "$0")")
echo "$temp_dir"
[rporresm@rporresm-mac test]$ bash example\ with\ spaces.sh
/var/folders/dx/y5_klhc1187gnzzmyb4pswb40000gn/T/tmp.buSMNjRKyQ-example with spaces.sh
```
I then remove the quote
```
[rporresm@rporresm-mac test]$ cat example\ with\ spaces.sh
temp_dir=$(mktemp -d --suffix "-$(basename $0)")
echo "$temp_dir"
[rporresm@rporresm-mac test]$ bash example\ with\ spaces.sh
basename: extra operand ‘spaces.sh’
Try 'basename --help' for more information.
/var/folders/dx/y5_klhc1187gnzzmyb4pswb40000gn/T/tmp.BugEWJylKo-
```

hack/build_channel_catalog.sh
rporres: Apologies in advance, but I don't understand what you mean here, @2uasimojo 😅  I got lost in _"In general, setting replaces to the latest skipped version"_ as it is something that we avoid with the `SKIP` logic

hack/build_channel_catalog.sh
2uasimojo: Whoah.

I'm having a hard time following how bash is parsing that. But your example is convincing. I'll go away now.

hack/build_channel_catalog.sh
2uasimojo: > Apologies in advance, but I don't understand what you mean here, @2uasimojo  I got lost in _"In general, setting replaces to the latest skipped version"_ as it is something that we avoid with the `SKIP` logic

Sorry, my "in general" was _too_ general.

I meant that, in the corner case where we need to skip all existing versions, we can set `replaces` to the latest skipped version. That's the only time `replaces` has a value covered by `skipRange`. This would be the correct thing to do for this corner case in The World (i.e. outside of appsre).

Specifically in appsre it doesn't matter, since we're purging the bad versions manually.

rrati: In the case we need to skip all existing versions, we just end up with a new catalog.  There's no need to set skip for our use case, and in fact setting skip/replaces for a catalog w/o the versions being skipped/replaced doesn't make sense and might cause problems.  I don't think this is something we need to worry about.  Current process will handle this case with a new catalog.  We only allow good version in a catalog.

rporres: Clarified. Let's keep `replaces`, and let's fix the edge case by adding in the replaces the last SKIP version in that very specific case to avoid manual intervention if possible.

rporres: fixed in 989a514

deploy/bundle/generate-csv-template.py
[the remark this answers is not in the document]
2uasimojo: Right; so using `replaces` *and* `skipRange` would work for both variants. Non-semver will observe `replaces` and `skips`/`skipRange`; but semver will ignore `replaces` and `skips` but observe `skipRange`. So the most generic and future-proof variant would be to use `replaces`+`skipRange`.

hack/build_channel_catalog.sh
2uasimojo: ✓

hack/build_channel_catalog.sh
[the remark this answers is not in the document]
2uasimojo: Okay, so can we include some richer information in this message? (Can be in a fup PR)

2uasimojo (commented):
I think this is in good shape. A couple of remaining things I would like to see addressed, but they can be done in a followon.

/lgtm

rrati:
@2uasimojo Can you explain why we need to keep bad operator versions in the catalog?  We are not running a real-world use case in CI/CD.  We also can't arbitrarily skip versions based upon the last CSV.  The original proposal was we would purge any bad versions of the operator from the catalog, and recovery would be to nuke the existing installs on the cluster.  I don't see a reason to deviate from the original implementation.  If an operator version is so bad that it can't ever be deployed, then I don't see the point in keeping it in the catalog at the expense of complicating the build process unnecessarily.

2uasimojo:
> @2uasimojo Can you explain why we need to keep bad operator versions in the catalog? We are not running a real-world use case in CI/CD. We also can't arbitrarily skip versions based upon the last CSV. The original proposal was we would purge any bad versions of the operator from the catalog, and recovery would be to nuke the existing installs on the cluster. I don't see a reason to deviate from the original implementation. If an operator version is so bad that it can't ever be deployed, then I don't see the point in keeping it in the catalog at the expense of complicating the build process unnecessarily.

I'm looking broader than just DVO. Thinking about the way we deploy the OSD operators, it _complicates_ the build process to require bad versions to be manually purged. And beyond that, we'd also like to accommodate third parties using our operators, at least in theory. In that scenario, keeping bad versions in the catalog as skipped allows OLM to (properly, generically) handle upgrades of clusters that are currently running the bad version.

rrati:
We are departing from the old practices to produce something new.  Purging old operators simplifies the build process by a lot because there is no need to keep track of good/current version or bad versions.  It's a simple grep operation to find the latest good operator and build the bundle/catalog off of that.

Likewise, the recovery, for all operators, needs to be a purge of the installed bad operator and re-install.  A major goal of this process change is to use the same upgrade chain in stage as in prod so that we can verify the upgrade chain.  Keeping bad operators in stage and upgrading from them violates that objective.  Do you have another way to accomplish this without purging deployed bad operators?

2uasimojo:
I'm having trouble understanding what we're debating here.

The process you describe will work fine no matter _what_ we put in the CSV, as you're effectively always asking OLM to install from fresh. If the operator isn't on the cluster, OLM doesn't even consult `replaces` or `skips` or `skipRange` -- it just installs the current version.

What I'm advocating is setting up those fields "correctly" anyway -- as if the *O*perator will, now or in the future, be deployed to a "real" cluster where its *L*ifecycle actually needs to be *M*anaged by *OLM* rather than by appsre.

I don't feel like this is overly complicated -- it's like a dozen lines of code in the script -- and it sounds like it'll be irrelevant to your environment anyway.

Am I missing something? Perhaps we should Meet™?

rrati:
After discussion it was decided to leave things as they are and use skipRange to denote bad versions that need to be skipped.  While we will never be upgrading from a bad version for our use case, if we were to use the same tooling to build and deploy the community release and push to community-operators then we would need to support skips/skipRange.  Having a common build process for both use cases is attractive.  Additionally since this is intended to go into boilerplate it is probably best to support the "right way" as best as possible to make the work in boilerplate as functional as possible.

rporres:
I've merged this as it is as I don't really understand why we need `skipRange` while we still maintain `replaces` since it looks like `skipRange` would only provide syntactic over `skips` and I definitely needed to move on. I discussed this with @2uasimojo on Friday and thought we had an agreement on it.

rporres:
Anyway the change should be pretty straight forward, so if you feel strongly about `skipRange`, please let me know and I will change it before starting to test this seriously.

rrati:
Since skipRange uses semver and skip does not, and semver seems to be the direction the community is heading we should use skipRange for futureproofing.

PR #60, opened 2020-11-17
Use skip range

rporres:
From #56 last comments

### DIFF
--- a/deploy/bundle/generate-csv-template.py
+++ b/deploy/bundle/generate-csv-template.py
@@ -58,10 +58,8 @@ if args.replaces:
     csv['spec']['replaces'] = f'deployment-validation-operator.v{args.replaces}'
 
 if args.skip:
-    csv['spec']['skips'] = [
-        f'deployment-validation-operator.v{version}' for version in
-        sorted(args.skip, key=lambda v: int(v.split('.')[2].split('-')[0]))
-    ]
+    csv['metadata']['annotations']['olm.skipRange'] = ' || '.join(
+        sorted(args.skip, key=lambda v: int(v.split('.')[2].split('-')[0])))
 
 now = datetime.datetime.now()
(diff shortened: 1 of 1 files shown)


rrati (approved):
/lgtm

deploy/bundle/generate-csv-template.py
```diff
@@ -58,10 +58,8 @@
     csv['spec']['replaces'] = f'deployment-validation-operator.v{args.replaces}'
 
 if args.skip:
-    csv['spec']['skips'] = [
-        f'deployment-validation-operator.v{version}' for version in
-        sorted(args.skip, key=lambda v: int(v.split('.')[2].split('-')[0]))
-    ]
+    csv['metadata']['annotations']['olm.skipRange'] = ' || '.join(
```
2uasimojo, line 61: ✓

rporres: They are the semver versions, not the full csv name. That's what we had before.

2uasimojo: Nice.

So... don't we also need to set the `version` field?

rrati: Yes.  Version field needs to be set for skipRange to work.

rporres: Version in the csv: https://github.com/app-sre/deployment-validation-operator/blob/master/deploy/bundle/template/deploymentvalidationoperator.clusterserviceversion.yaml#L63

And it's passed via `oc process` https://github.com/app-sre/deployment-validation-operator/blob/master/hack/build_channel_catalog.sh#L136

2uasimojo: Ah, cool, I didn't look far enough to see that bit. Thanks!

rporres: yw!

PR #67, opened 2020-11-23
Remove dvo specific logic in bundle creation

rporres:
* Python script now has all the logic to create bundle contents, we don't use `oc` anymore.
* We use opm to build the bunde to remove specific dockerfiles from the repository
* Remove exports of all variables as this may not what we will find in boilerplate
* With this we're very close to use this script from [`boilerplate`](https://github.com/openshift/boilerplate/):
  * We will need to move the commands installation to [`ensure.sh`] if we find a good way to do it (https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/ensure.sh)
  * Python script to generate bundle contents will still be needed everywhere until we establish a standard in manifests location and content.


rporres:
cc @2uasimojo @rrati

Makefile
```diff
@@ -43,4 +37,19 @@ docker-login:
 
 .PHONY: catalog
 channel-catalog: docker-login
-	@./hack/build_channel_catalog.sh
+	@BUNDLE_IMAGE="${BUNDLE_IMAGE}" \
```
2uasimojo, line 40: Personally, I don't like `@`. I want to see the CLI that was invoked -- especially in a case like this with all the variable settings, considering how significant they are to the process.

hack/build_channel_catalog.sh
```diff
@@ -129,37 +176,28 @@ if [[ ${#skip_versions[@]} -gt 0 ]]; then
...
 # TODO: remove this tag if we're in dry-run mode in the cleanup trap script
 [[ "$DRY_RUN" == "false" ]] || bundle_image_current_commit="${bundle_image_current_commit}-dryrun"
 log "Creating bundle image $bundle_image_current_commit"
-
-# TODO: use opm to build this
-$CONTAINER_ENGINE build -t "$bundle_image_current_commit" "$BUNDLE_DEPLOY_DIR"
+pushd "$bundle_temp_dir"
+opm alpha bundle build --directory "$manifests_temp_dir" \
+                       --channels alpha \
+                       --default alpha \
```
2uasimojo, line 195: Will this be `alpha` for everyone?

hack/generate-operator-bundle-contents.py
```diff
@@ -1,35 +1,47 @@
...
 import argparse
 
-parser = argparse.ArgumentParser()
-parser.add_argument("-r", "--replaces", help="Replaces version", type=str)
-parser.add_argument('-s','--skip', action='append',
+parser = argparse.ArgumentParser(add_help=False)
+required = parser.add_argument_group('required arguments')
+
+required.add_argument('-n', '--name', help='operator name', type=str, required=True)
+required.add_argument('-v', '--version', help='operator version', type=str, required=True)
```
2uasimojo, line 16: nit: `--version` makes me think I'm going to get the version of the tool.
Since we're dealing with multiple "version" args (`replaces`, `skip`), perhaps this could be `-c`/`--current-version`?

This is totally not a big deal though.

hack/generate-operator-bundle-contents.py
```diff
@@ -1,35 +1,47 @@
...
+optional.add_argument( '-h', '--help', action='help', default=argparse.SUPPRESS,
+    help='show this help message and exit')
+optional.add_argument('-r', '--replaces', help='Replaces version', type=str)
+optional.add_argument('-s','--skip', action='append',
         help='Skips version (can be specified multiple times)')
+
 args = parser.parse_args()
 
-root = pathlib.Path(__file__).parent.absolute() / '../..'
+root = pathlib.Path(__file__).parent.absolute() / '..'
```
2uasimojo, line 30: TIL :)

hack/build_channel_catalog.sh
```diff
@@ -129,37 +176,28 @@ if [[ ${#skip_versions[@]} -gt 0 ]]; then
...
-    log "Using local opm version $opm_local_executable"
-else
-    opm_download_url="https://github.com/operator-framework/operator-registry/releases/download/$OPM_VERSION/${GOOS}-${GOARCH}-opm"
-    log "Downloading opm from $opm_download_url to $temp_dir/opm"
-    curl -s -L "$opm_download_url" -o "$temp_dir/opm"
-    chmod u+x "$temp_dir/opm"
-    opm_local_executable="$temp_dir/opm"
-fi
+manifests_temp_dir=$(mktemp -d -p "$bundle_temp_dir" manifests.XXXX)
+./hack/generate-operator-bundle-contents.py --name "$OPERATOR_NAME" \
```
2uasimojo, line 180: This assumes the script is invoked from the repo root, right?

To be more explicit, we could compute the directory of this script and use that instead.

```
HERE=${0%/*}
...
$HERE/generate-operator-bundle-contents.py ...
```

deploy/bundle/template/deploymentvalidationoperator.clusterserviceversion.yaml
```diff
@@ -1,71 +1,57 @@
...
-    - supported: false
-      type: MultiNamespace
-    keywords:
-    - dvo
-    labels:
+  annotations:
+    capabilities: Basic Install
+    categories: Application Runtime, Monitoring, Security
+    certified: "false"
+    # containerImage: ${IMAGE}:${IMAGE_TAG}
```
2uasimojo, line 9: Are these comments left in deliberately? (throughout)

hack/generate-operator-bundle-contents.py
```diff
@@ -1,35 +1,47 @@
...
 manifest_dir = root / 'deploy/openshift'
 csv_template_dir = root / 'deploy/bundle/template'
 
 with open(csv_template_dir / 'deploymentvalidationoperator.clusterserviceversion.yaml',
```
2uasimojo, line 34: Should this use some form of `args.name`?

(Perhaps it would be simplest to expect the name of the template to include the hyphens. So in this case `deployment-validation-operator.clusterserviceversion.yaml`.)

Makefile
rporres: Fine by me :)

hack/build_channel_catalog.sh
rporres: Yes. I can move that into a variable, but the fact remains that we will only use one channel and this has to match whatever is in the subscription, e.g. https://github.com/app-sre/deployment-validation-operator/blob/master/deploy/openshift/deployment-validation-operator-olm.yaml#L21

hack/generate-operator-bundle-contents.py
rporres: I like the idea. I will change it.

hack/build_channel_catalog.sh
rporres: I assumed that since this is invoked through a `Makefile` target... I can try with the `HERE` approach :)

deploy/bundle/template/deploymentvalidationoperator.clusterserviceversion.yaml
rporres: Yes, those are the fields that will be filled by the python script

hack/generate-operator-bundle-contents.py
rporres: totally I will change it

Makefile
rporres: Fixed in 5a804b5

hack/generate-operator-bundle-contents.py
rporres: fixed in 7415dc6

hack/generate-operator-bundle-contents.py
rporres: fixed in 9f2a90b

hack/build_channel_catalog.sh
rporres: fixed in 903ab58

hack/build_channel_catalog.sh
rporres: fixed in eb45b68

deploy/bundle/template/deploymentvalidationoperator.clusterserviceversion.yaml
rporres: added note in 5a61422

2uasimojo:
Thanks @rporres.

/lgtm

PR #70, opened 2020-11-23
Fix bug in opm cmd name when building bundle

rporres:


2uasimojo (commented):
Oh. Yeah.

/lgtm

PR #84, opened 2021-02-12
Don't check for CPU limits

cblecker:
There are [volumes](https://www.google.com/search?q=don%27t+use+cpu+limits) written on this, but setting CPU limits is generally a bad idea for production workloads. It can drastically slow applications down, because of the way CFS shares are sliced and calculated.

Requests are great for both CPU and memory as they aid in scheduling. Memory limits can serve as a safety mechanism for systems without swap, and are immediately actionable by the OOMKiller. But CPU limits in the CFS scheduler are hard to set, and they can slow your application down by restricting CPU time, even if you aren't anywhere close to the limit set.

This PR removes limits from this metric.

cc @jewzaam @mwoodson @jharrington22 @robotmaxtron FYI

### DIFF
--- a/pkg/testutils/templates/Deployment.yaml.tpl
+++ b/pkg/testutils/templates/Deployment.yaml.tpl
@@ -115,7 +115,6 @@ spec:
         resources:
           {{- if .ResourceLimits }}
           limits:
-            cpu: "1"
             memory: 1Gi
           {{- end }}
           {{- if .ResourceRequests }}
--- a/pkg/testutils/templates/ReplicaSet.yaml.tpl
+++ b/pkg/testutils/templates/ReplicaSet.yaml.tpl
@@ -118,7 +118,6 @@ spec:
         resources:
           {{- if .ResourceLimits }}
           limits:
-            cpu: "1"
             memory: 1Gi
           {{- end }}
           {{- if .ResourceRequests }}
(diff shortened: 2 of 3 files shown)


rporres:
Fixing validate job in https://github.com/app-sre/deployment-validation-operator/pull/85

jmelis:
@cblecker thanks for this proposal. Right now we don't have consensus within AppSRE whether we would like to stop enforcing this across all our multitenant clusters.

While we understand that this may cause applications to underperform, we can monitor it with the cpu throttling metrics. If we remove this, we may begin start seeing bad neighbour effects, where some applications may take down other applications.

That said, we do want to start shifting towards this model where we don't enforce defining cpu limits, but we would like to do this gradually and in controlled environments.

We are currently working on adding configmap support to this operator, so validations can be enabled or disabled on a per namespace basis, and more importantly, it will allow us to parametrize checks.

I see this as follows:

```yaml
config.yaml: |
    requests_limits:
        enabled: true
        parameters:
            cpu_limit: false
```

That will allow us to keep by default your proposal (no check) but provides us with a knob that enables this check for our multi-tenant namespace?

What do you think?

cblecker:
@jmelis I am completely fine with it being an optional parameter.. or just exposing metrics for all four, and then tuning the alerts to what you need. My main concern is by default, we shouldn't be pushing for CPU limits (in particular if we want to turn this on in managed environments like OSD)

rrati:
This will likely need to be addressed in the kube-linter check now.  Closing.

PR #86, opened 2021-03-23
Do not wrap main in or condition

rporres:

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
+++ b/boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
@@ -379,5 +379,5 @@ function main() {
 
 # Main
 if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
-  main || exit 1
+  main
 fi


rporres:
It is needed because `set -e` does not affect conditions. I found a great way to deactivate it completely 😂  As an example, if I execute this code 
```
set -e

function fail() {
    ls /noo
}

fail || echo "adios"
echo "hola"
```
I get
```
ls: cannot access '/noo': No such file or directory
adios
hola
```

2uasimojo:
That's because `echo "adios"` succeeds. Change that to `exit 1` and `hola` is never printed.

2uasimojo:
Though interestingly, whereas `ls /bar || ls /baz` triggers immediate failure after running both sides, `ls /bar && ls /baz` does not trigger immediate failure, which I find weird.

Anyway, it's a fine change -- I assume you intend to propose it to boilerplate.

rporres:
> That's because echo "adios" succeeds. Change that to exit 1 and hola is never printed.

My point is that I've managed to deactivate the `set -e`. I guess I didn't explain well what I have in mind. Take a look to this script:

```
set -e

function fail() {
    ls /noo
    echo "good"
}

fail || { echo "I've failed"; exit 1; }
echo "I'm still alive"
```
which executes as
```
ls: cannot access '/noo': No such file or directory
good
I'm still alive
```
whereas if i don't check immediately with an or
```
set -e

function fail() {
    ls /noo
    echo "good"
}

fail
echo "I'm still alive"
```
it fails immediately
```
ls: cannot access '/noo': No such file or directory
```

rporres:
> Anyway, it's a fine change -- I assume you intend to propose it to boilerplate.

Yes, I want it to go through a real pipeline before adding it to boilerplate

PR #93, opened 2021-05-24
Use kube-linter as check engine

rrati:


Makefile
```diff
@@ -18,6 +18,9 @@ docker-test:
 	${CONTAINER_ENGINE} build . -f $(OPERATOR_DOCKERFILE).test -t $(OPERATOR_IMAGE_URI_TEST)
 	${CONTAINER_ENGINE} run -t $(OPERATOR_IMAGE_URI_TEST)
 
+generated-srcs:
```
2uasimojo, line 21: Unless there's some reason you need a target of this exact name, boilerplate [already has](https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/standard.mk#L137) a `go-generate` target for this

go.mod
```diff
@@ -3,26 +3,49 @@ module github.com/app-sre/deployment-validation-operator
 go 1.13
```
2uasimojo, line 3: Consider
```suggestion
go 1.16
```
Not because it actually does anything, but as a hint to contributors, since you are actually relying on 1.16 for $reasons.

2uasimojo (commented):
Are *all* of these changes related to this kube-linter thing? Looked like others (config file processing?) might have snuck in.

go.mod
rrati: Yep, this I meant to do and forgot.  Thanks for catching.

Makefile
rrati: Cool, will use that.

PR #94, opened 2021-05-26
Use kube-linter as check engine

rrati:
Using kube-linter as the check engine caused a number of significant changes to the code base:

- A new version of go was required to work with embed (1.16)
- Updates to numerous dependencies, including operator-sdk and controller-runtime
   - This in turn caused changes to things like the metrics service.  The package used before in operator-sdk became an internal package and unable by importers.  The metrics service/service monitor is now expected to be handled via yaml external to the operator
- In order to use the kube-linter engine, we have to create and initialize its the check registry.  This loads the check definition yaml and configures the templates which are stored in the check registry.  The code to do this is largely copied from kube-linter, but since kube-linter does this work in a corba command we can't really re-use it sanely.
- The replicas check has been re-written to work with the kube-linter check definition and the yaml for check definition is in the configuration file passed to dvo/kube-linter as a custom check.  This will likely change in the future.
- Previously the checks themselves handled their own metrics, but kube-linter doesn't do metrics.  It runs an object through checks that support that kind of object and returns a report of checks that fail.  This means the method running the checks has to also handle processing the results and generating the metrics.  This implementation currently has a problem with flapping metrics that will be addressed later.

kube-linter itself defines checks in 2 parts, a template and a yaml definition of a check.  The template is the code that implements a check and allows specific variables to configure it.  The yaml definition of a check calls out what template to use, what object kinds to pass to the check, and what variables to use for that check.  The yaml definition is what controls what kube-linter will run against objects.


pkg/controller/controller.go
```diff
@@ -24,3 +24,74 @@ func AddToManager(m manager.Manager) error {
...
+	for i := range kubeLinterKinds {
+		if kubeLinterKinds[i] == objectkinds.Any {
+			kubeLinterKinds = append(kubeLinterKinds[:i], kubeLinterKinds[i+1:]...)
+			break
+		}
+	}
+
+	kubeLinterMatcher, err := objectkinds.ConstructMatcher(kubeLinterKinds...)
+	if err != nil {
+		// TODO log error or exit?
```
rporres, line 42: what kind of errors will we have here?

rrati: This should never return an error unless there is a problem in kube-linter itself afau

rporres (commented):
Using kube-linter is definitely the way to go, so this looks great

I would state clearly which files are copied / translated / inspired and the way ro remove them as it is done inconsistently.

rporres:
In general, some general doc on how the integration with kube-linter works and the boundaries would be great.

PR #97, opened 2021-06-02
Delete metrics for validations not in the report from kube-linter

rrati:
The problem to solve is two-fold:
1. Original implementation was setting 0 to clear metrics, generating a LOT of metrics for passed validations
2. Original implementation was resetting all labels for a metric before processing the results.  This would cause metric flapping for failed validations

This change attempts to handle the above intelligently.  When clearing the metrics, it will delete the labels from a metric if that metric is not in the list of failed checks returned by kube-linter.  Then the processing of the results continues as before, setting a metric for each failed validation.

### DIFF
--- a/pkg/validations/validation_engine.go
+++ b/pkg/validations/validation_engine.go
@@ -13,6 +13,7 @@ import (
 	"golang.stackrox.io/kube-linter/pkg/checkregistry"
 	"golang.stackrox.io/kube-linter/pkg/config"
 	"golang.stackrox.io/kube-linter/pkg/configresolver"
+	"golang.stackrox.io/kube-linter/pkg/diagnostic"
 
 	// Import and initialize all check templates from kube-linter
 	_ "golang.stackrox.io/kube-linter/pkg/templates/all"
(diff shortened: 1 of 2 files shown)


pkg/validations/validation_engine.go
```diff
@@ -105,9 +110,19 @@ func (ve *validationEngine) DeleteMetrics(labels prometheus.Labels) {
 	}
 }
 
-func (ve *validationEngine) ClearMetrics(labels prometheus.Labels) {
-	for _, v := range ve.metrics {
-		v.With(labels).Set(0)
+func (ve *validationEngine) ClearMetrics(reports []diagnostic.WithContext, labels prometheus.Labels) {
+	// Create a list of validation names for use to delete the labels from any
```
rporres, line 114:
```suggestion
	// Create a set of validation names for use to delete the labels from any
```

rporres (approved):
Since there's no difference between a metric set to `0` or non-existing, it is desirable to not emit `0` metrics to make Prometheus life easier

PR #100, opened 2021-09-28
only ignore deployment-like resources owned by deployment-like resources

BumbleFeng:
Signed-off-by: Feng Huang <fehuang@redhat.com>

### DIFF
--- a/pkg/utils/object.go
+++ b/pkg/utils/object.go
@@ -1,8 +1,13 @@
 package utils
 
 import (
+	"golang.stackrox.io/kube-linter/pkg/objectkinds"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+var deploymentLikeMatcher, _ = objectkinds.ConstructMatcher(objectkinds.DeploymentLike)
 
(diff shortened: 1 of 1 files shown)


pkg/utils/object.go
```diff
@@ -1,18 +1,33 @@
...
+// IsOwner returns false if the object is deployment-like resource and owned by a deployment-like resource
+func IsOwner(obj client.Object) bool {
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	// no need to check owner for none deployment-like resource
+	if deploymentLikeMatcher.Matches(gvk) {
+		for _, ref := range obj.GetOwnerReferences() {
+			refGvk := schema.FromAPIVersionAndKind(ref.APIVersion, ref.Kind)
+			// get a deployment-like owner
+			if deploymentLikeMatcher.Matches(refGvk) {
+				return false
```
rrati, line 28: Won't this still follow the chain all the way to the operator and not stop at an operand?  If i have pod worker1 created by deployment worker1 that was created by pod operator1 created by deployment operator1, then this would end up reporting only on deployment operator1 wouldn't it?

BumbleFeng: The owner of deployment worker1 will be CSV instead of the pod operator1.

rrati: Makes life easier then.  Excellent.

rrati (approved):
/lgtm

PR #102, opened 2021-09-29
Disable incompatible checks

bkez322:
This will force DVO to disable kube-linter checks that don't work correctly in DVO. Also adds a test.

### DIFF
--- a/pkg/validations/base_test.go
+++ b/pkg/validations/base_test.go
@@ -1,6 +1,7 @@
 package validations
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/app-sre/deployment-validation-operator/pkg/testutils"
@@ -162,14 +163,12 @@ func getAllBuiltInKubeLinterChecks() ([]string, error) {
 	}
 	registry := checkregistry.New()
 	if err := builtinchecks.LoadInto(registry); err != nil {
-		log.Error(err, "failed to load built-in validations")
-		return nil, err
+		return nil, fmt.Errorf("failed to load built-in validations: %s", err.Error())
 	}
 
 	enabledChecks, err := configresolver.GetEnabledChecksAndValidate(&ve.config, registry)
 	if err != nil {
-		log.Error(err, "error finding enabled validations")
-		return nil, err
+		return nil, fmt.Errorf("error finding enabled validations: %s", err.Error())
 	}
 
 	return enabledChecks, nil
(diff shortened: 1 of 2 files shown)


pkg/validations/base_test.go
```diff
@@ -109,3 +129,34 @@ func TestRunValidationsIssueCorrection(t *testing.T) {
...
 }
+
+func TestIncompatibleChecksAreDisabled(t *testing.T) {
+	e, err := createEngineWithAllChecks()
+	if err != nil {
+		t.Errorf("Error creating validation engine %v", err)
+	}
+
+	enabledChecks := e.EnabledChecks()
+	if len(enabledChecks) <= 10 {
```
rrati, line 140: This will change as we add checks.  Can we dynamically determine what the expected number of enabled checks is?

bkez322: The current number is actually 37 checks (excluding the disabled checks). '10' was an arbitrary number I chose to make sure we weren't loading an empty list or a list from the other test.

With that said, I'll look at rewriting this piece.

pkg/validations/base_test.go
```diff
@@ -55,6 +56,25 @@ func createEngine() (validationEngine, error) {
 	return ve, nil
 }
 
+func createEngineWithAllChecks() (validationEngine, error) {
```
rrati, line 59: Can we combine this with previous function?  The differences seem like that can be passed in via function args.

rrati: Not sure we need to keep the sync.Once in these tests.

bkez322: >Can we combine this with previous function? The differences seem like that can be passed in via function args.

I'll look at doing that.

>Not sure we need to keep the sync.Once in these tests.

Yeah probably not. It doesn't appear to be functionally doing anything, but I didn't touch it in case it was added for a reason I didn't understand. I'll go ahead and remove it.

pkg/validations/base_test.go
bkez322: Resolved with:
https://github.com/app-sre/deployment-validation-operator/blob/8fe3c3289499cb0c2ec8433920c0b7ccbb7e2551/pkg/validations/base_test.go#L128-L139

pkg/validations/base_test.go
bkez322: I removed use the sync package, and I refactored the engine construction functions to be more modular:

https://github.com/app-sre/deployment-validation-operator/blob/8fe3c3289499cb0c2ec8433920c0b7ccbb7e2551/pkg/validations/base_test.go#L27-L65

rrati (approved):
/lgtm

PR #105, opened 2021-10-01
Updated to README and deployment yamls

rrati:


BumbleFeng:
LGTM

you can also delete https://github.com/app-sre/deployment-validation-operator/blob/master/deploy/observability/service.yaml

PR #118, opened 2021-11-01
Check to clear metrics if 0 replicas

npecka:
Added condition where no validations are run and metrics cleared if object is found to have replica count of 0

### DIFF
--- a/pkg/validations/base.go
+++ b/pkg/validations/base.go
@@ -45,12 +45,14 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 	spec := objValue.FieldByName("Spec")
 	if spec.IsValid() {
 		replicas := spec.FieldByName("Replicas")
-		numReplicas, ok := replicas.Interface().(*int32)
+		if replicas.IsValid() {
+			numReplicas, ok := replicas.Interface().(*int32)
 
-		// clear labels if we fail to get a value for numReplicas, or if value is <= 0
-		if !ok || numReplicas == nil || *numReplicas <= 0 {
-			engine.DeleteMetrics(promLabels)
-			return
+			// clear labels if we fail to get a value for numReplicas, or if value is <= 0
+			if !ok || numReplicas == nil || *numReplicas <= 0 {
+				engine.DeleteMetrics(promLabels)
+				return
+			}
 		}
 	}
 
--- a/pkg/validations/base.go
+++ b/pkg/validations/base.go
@@ -46,7 +46,9 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 	if spec.IsValid() {
 		replicas := spec.FieldByName("Replicas")
 		numReplicas, ok := replicas.Interface().(*int32)
-		if ok && numReplicas != nil && *numReplicas <= 0 {
+
+		// clear labels if we fail to get a value for numReplicas, or if value is <= 0
+		if !ok || numReplicas == nil || *numReplicas <= 0 {
 			engine.DeleteMetrics(promLabels)
 			return
 		}
--- a/pkg/validations/base.go
+++ b/pkg/validations/base.go
@@ -36,11 +36,6 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 	lintCtx := &lintContextImpl{}
 	lintCtx.addObjects(lintcontext.Object{K8sObject: obj})
 	lintCtxs = append(lintCtxs, lintCtx)
-	result, err := run.Run(lintCtxs, engine.CheckRegistry(), engine.EnabledChecks())
-	if err != nil {
-		log.Error(err, "error running validations")
-		return
-	}
 
 	// If controller has no replicas clear existing metrics and
 	// do not run any validations
@@ -49,6 +44,12 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 		engine.DeleteMetrics(promLabels)
 		return nil
 	}
+
+    result, err := run.Run(lintCtxs, engine.CheckRegistry(), engine.EnabledChecks())
+    if err != nil {
+    	log.Error(err, "error running validations")
+    	return
+    }
 
 	// Clear labels from past run to ensure only results from this run
 	// are reflected in the metrics
--- a/pkg/validations/base.go
+++ b/pkg/validations/base.go
@@ -41,11 +41,9 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 	if spec.IsValid() {
 		replicas := spec.FieldByName("Replicas")
 		numReplicas, ok := replicas.Interface().(*int32)
-		if ok {
-			if *numReplicas <= 0 {
-				engine.DeleteMetrics(promLabels)
-				return
-			}
+		if ok && numReplicas != nil && *numReplicas <= 0 {
+			engine.DeleteMetrics(promLabels)
+			return
 		}
 	}
 
--- a/pkg/validations/base.go
+++ b/pkg/validations/base.go
@@ -40,11 +40,24 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 	spec := objValue.FieldByName("Spec")
 	if spec.IsValid() {
 		replicas := spec.FieldByName("Replicas")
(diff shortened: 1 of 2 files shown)


pkg/validations/base.go
```diff
@@ -31,6 +34,18 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 		return
 	}
 
+	// If controller has no replicas clear existing metrics and
+	// do not run any validations
+
+	replicas, err := yaml.Marshal(obj)
```
rrati, line 40: This isn't necessary.  Would be better to use a reflect methodology similar to extract.Replicas from kube-linter.  Taking a Go object and transforming it to string to check and field isn't the right path.

pkg/validations/base.go
```diff
@@ -31,6 +33,32 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
...
+					if *numReplicas <= 0 {
+						engine.DeleteMetrics(promLabels)
+						return
+					}
+				}
+			}
+		}
+	}
+
+	//numReplicas, ok := replicas.Interface().(*int32)
```
rrati, line 56: Remove commented out code

pkg/validations/base.go
```diff
@@ -31,6 +33,32 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
...
+	// If controller has no replicas clear existing metrics and
+	// do not run any validations
+
+	objValue := reflect.Indirect(reflect.ValueOf(obj))
+	spec := objValue.FieldByName("Spec")
+	if spec.IsValid() {
+		replicas := spec.FieldByName("Replicas")
+		if replicas.IsValid() {
+			numReplicas, ok := replicas.Interface().(*int32)
+			if ok {
```
rrati, line 45: Once you've gotten here, you know you have an object with a valid replicas field, so any failures should produce a log message of some kind

pkg/validations/base.go
```diff
@@ -31,6 +33,32 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
...
+				if numReplicas != nil {
```
rrati, line 46: This technically should never happen.  I'm not sure what condition could result in a nil replicas field.  I would just treat this as if replicas are 0.  You can compress this to:

if numReplicas ** *numReplicas <= 0 {
...
} else {
...
}

rrati (changes requested):

pkg/validations/base.go
npecka: @rrati Using the change I receive an error saying `cannot indirect *numReplicas`. Since you mentioned there isn't a case where the result could be nil at that point should we just leave it as `if *numReplicas <= 0` at that point and omit the left side of the statement all together?

rrati: > > `if numReplicas ** *numReplicas <= 0`
> 
> Is this supposed to say: `if numReplicas == nil || *numReplicas <= 0` ?

This was supposed to be `numReplicas && *numReplicas <= 0`

rrati: > @rrati Using the change I receive an error saying `cannot indirect *numReplicas`. Since you mentioned there isn't a case where the result could be nil at that point should we just leave it as `if *numReplicas <= 0` at that point and omit the left side of the statement all together?

I'm not aware of an instance where this could be nil.  I tend to favor good defensive programming through and checking that a ptr is not nil should be done before acting upon it.

npecka: @rrati Does the current change work? numReplicas is an int32 so needs to be compared to something for it to work. I condensed the if statement. Might not need the `ok` checked with all of that if numReplicas validity is checked.

bkez322: Considering this suggestion from Rob: 
> I'm not sure what condition could result in a nil replicas field. I would just treat this as if replicas are 0.

I think this implies we want to try to clear labels if either of the following cases are true:
* "replicas" field doesn't exist (we fail to get a value for numReplicas)
* "replicas" field exists, and is 0 or less

Therefore I propose this block of code should look like this:
```
objValue := reflect.Indirect(reflect.ValueOf(obj))
spec := objValue.FieldByName("Spec")
if spec.IsValid() {
	replicas := spec.FieldByName("Replicas")
	numReplicas, ok := replicas.Interface().(*int32)

        // clear labels if we fail get a value for numReplicas, or if value is 0 or less
	if !ok || numReplicas == nil || *numReplicas <= 0 { 
		engine.DeleteMetrics(promLabels)
		return
	}
}
```
or maybe this (whichever you find more readable, they should both do the same thing):
```
objValue := reflect.Indirect(reflect.ValueOf(obj))
spec := objValue.FieldByName("Spec")
if spec.IsValid() {
	replicas := spec.FieldByName("Replicas")
	numReplicas, ok := replicas.Interface().(*int32)

        // treat missing numReplicas value as 0
        if !ok || numReplicas == nil {
            val := 0
            *numReplicas = &val
        }

	if *numReplicas <= 0 { 
	   engine.DeleteMetrics(promLabels)
	   return
	}
}
```

pkg/validations/base.go
```diff
@@ -37,6 +38,22 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 		return
 	}
 
+	// If controller has no replicas clear existing metrics and
+	// do not run any validations
+
+	objValue := reflect.Indirect(reflect.ValueOf(obj))
+	spec := objValue.FieldByName("Spec")
+	if spec.IsValid() {
+		replicas := spec.FieldByName("Replicas")
```
rrati, line 47: Need to verify that replicas is valid.  Spec field can exist on an object w/o replicas existing in the spec

pkg/validations/base.go
```diff
@@ -37,6 +38,24 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
...
+		if replicas.IsValid() {
+			numReplicas, ok := replicas.Interface().(*int32)
+
+			// clear labels if we fail to get a value for numReplicas, or if value is <= 0
+			if !ok || numReplicas == nil || *numReplicas <= 0 {
+				engine.DeleteMetrics(promLabels)
+				return
+			}
+		}
+	}
```
bkez322, line 57: @rrati

So is the idea here to just assume that objects with a `replicas` field are controllers?

Or do we want to explicitly check the object's kind to be a deployment, replicaset, statefulset, deploymentconfig, etc?

I think the former is probably fine, but hypothetically there may be k8s kinds with a `replicas` field that aren't controllers.

bkez322:
Asked a question about the assumption we're making with deployments and the `replicas` field.

If we plan on assuming that objects with a `replicas` field are controllers, then this PR LGTM. Thanks for adding a test!

pkg/validations/base.go
bkez322: A concern:

With the code as-is, I think someone could add a bogus `replicas` field  with value `0` to a `Service`, or a `ServiceAccount` (for example), and then bypass any checks.

bkez322: Hm, I just tested this, and k8s will just ignore invalid properties (e.g. `replicas`) in the stored YAML definition of applied objects. So this actually may not be a concern.

rrati: I don't know of any API objects that have a replicas field which are not controllers.  I think it's safe to assume that is the case unless we find an edge case otherwise.  It's possible that maybe a custom resource of some kind could have a replicas field that isn't a controller, but until we hit such a thing I don't think we should plan for it.

pkg/validations/base_test.go
```diff
@@ -202,6 +203,79 @@ func TestIncompatibleChecksAreDisabled(t *testing.T) {
...
+		t.Errorf("Error getting prometheus metric: %v", err)
+	}
+
+	// If metrics are cleared then the value will be == 0
+	if metricValue := int(prom_tu.ToFloat64(metric)); metricValue != 0 {
+		t.Errorf("Deployment test failed %#v: got %d want %d", customCheck.Name, metricValue, 0)
+	}
+
+	// Check using a replica count above 0
+	replicaCnt = int32(3)
```
rrati, line 248: Setting 3 replicas will cause the check to pass and no metric would be generated.  I would use 1 here so the check will fail and create a metric.

rrati (approved):
/lgtm

PR #119, opened 2021-11-03
"check_description" and "check_remediation" labels for metrics

bkez322:
(DVO-12)

Adds "check_description" and "check_remediation" as constant labels for check-validation metrics.

Prints out "check_description", "check_remediation", and "check_failure_reason" in logs.

Changes the "Help" string for metrics to include both a description and remediation.

Modifies a test to validate that metrics include the constant labels.

The validationEngine struct now stores `map[string]config.Check` for all registered checks, because I couldn't find any better way for `RunValidations()` to get access to kube-linter check descriptions.

### DIFF
--- a/pkg/validations/base.go
+++ b/pkg/validations/base.go
@@ -25,10 +25,7 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 	// If the object was deleted, then just delete the metrics and
 	// do not run any validations
 	if isDeleted {
-		for _, registeredCheck := range engine.registeredChecks {
-			fullPromLabelsForCheck := getFullPromLabels(basePromLabels, registeredCheck)
-			engine.DeleteMetrics(fullPromLabelsForCheck)
-		}
+		engine.DeleteMetrics(basePromLabels)
 		return
 	}
 
--- a/pkg/validations/base.go
+++ b/pkg/validations/base.go
@@ -50,8 +50,8 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 	engine.ClearMetrics(result.Reports, promLabels)
 
 	for _, report := range result.Reports {
-		check, err := engine.getCheckByName(report.Check)
-		if err != nil {
+		check, ok := engine.registeredChecks[report.Check]
+		if !ok {
 			log.Error(err, "Failed to get check '%s' by name", report.Check)
 			return
 		}
--- a/pkg/validations/base.go
+++ b/pkg/validations/base.go
@@ -50,19 +50,17 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 	engine.ClearMetrics(result.Reports, promLabels)
 
 	for _, report := range result.Reports {
-		checkDescription := ""
 		check, err := engine.getCheckByName(report.Check)
 		if err != nil {
 			log.Error(err, "Failed to get check '%s' by name", report.Check)
 			return
 		}
-		checkDescription = check.Description
 		logger := log.WithValues(
 			"request.namespace", request.Namespace,
 			"request.name", request.Name,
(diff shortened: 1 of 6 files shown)


pkg/validations/base_test.go
```diff
@@ -35,17 +36,24 @@ func newEngine(c config.Config) (validationEngine, error) {
 	return ve, nil
 }
 
+func newCustomCheck() config.Check {
+	return config.Check{
+		Name:        checkName,
+		Description: "some description",
+		Remediation: "some remediation",
```
rrati, line 43: I would make any string we are going to check a constant/variable so if it is changed, it will be changed everywhere that cares.

pkg/validations/validation_engine.go
```diff
@@ -97,11 +98,17 @@ func (ve *validationEngine) InitRegistry() error {
 		if check == nil {
 			return fmt.Errorf("unable to create metric for check %s", checkName)
 		}
-		metric := newGaugeVecMetric(strings.ReplaceAll(check.Spec.Name, "-", "_"),
-			// Should this be the Remediation text or the description?
-			// For now go with Description
-			check.Spec.Description,
-			[]string{"namespace", "name", "kind"})
+		ve.registeredChecks = append(ve.registeredChecks, check.Spec)
```
rrati, line 101: If the only use for this is to provide check information by name, and since a check name has to be unique, why not may this a map?

rrati (commented):
Looks good.  Few minor comments

pkg/validations/base_test.go
bkez322: https://github.com/app-sre/deployment-validation-operator/blob/3553f14b6071e509f4e3ba43a13ef4f70746af76/pkg/validations/base_test.go#L24-L29

https://github.com/app-sre/deployment-validation-operator/blob/3553f14b6071e509f4e3ba43a13ef4f70746af76/pkg/validations/base_test.go#L42-L53

https://github.com/app-sre/deployment-validation-operator/blob/3553f14b6071e509f4e3ba43a13ef4f70746af76/pkg/validations/base_test.go#L88-L124

I amended how these strings are handled.

pkg/validations/validation_engine.go
bkez322: https://github.com/app-sre/deployment-validation-operator/blob/a3f95f2cfcf310ec98d3cb9e15c7e0c0051734ff/pkg/validations/validation_engine.go#L31-L37

https://github.com/app-sre/deployment-validation-operator/blob/a3f95f2cfcf310ec98d3cb9e15c7e0c0051734ff/pkg/validations/base.go#L52-L57

No idea why I didn't make this a map. Good idea. Updated now.

pkg/validations/base.go
```diff
@@ -46,11 +50,20 @@ func RunValidations(request reconcile.Request, obj client.Object, kind string, i
 	engine.ClearMetrics(result.Reports, promLabels)
 
 	for _, report := range result.Reports {
+		check, ok := engine.registeredChecks[report.Check]
```
rrati, line 53: This should be a function.  engine.GetCheckByName or something.  The fact that the engine stores this as a map should be transparent external to the engine.

rrati (changes requested):

pkg/validations/base.go
bkez322: https://github.com/app-sre/deployment-validation-operator/blob/541effb687e5b5435c76b3eec278e7eb1d786436/pkg/validations/base.go#L55-L59

https://github.com/app-sre/deployment-validation-operator/blob/541effb687e5b5435c76b3eec278e7eb1d786436/pkg/validations/validation_engine.go#L174-L180

Done

rrati (approved):
/lgtm

PR #120, opened 2021-11-08
Fix DVO checks against itself (dvo-22)

bkez322:
This should fix DVO flagging itself with any failing checks.

Also fixes some apparent indent issues in `deploy/openshift/network-policies.yaml`.

### DIFF
--- a/deploy/openshift/operator.yaml
+++ b/deploy/openshift/operator.yaml
@@ -60,7 +60,7 @@ spec:
         - name: dvo-config
           mountPath: /config
         securityContext:
-          runAsUser: 1000
+          runAsNonRoot: true
           readOnlyRootFilesystem: true
       volumes:
       - name: dvo-config
--- a/deploy/openshift/operator.yaml
+++ b/deploy/openshift/operator.yaml
@@ -60,7 +60,6 @@ spec:
         - name: dvo-config
           mountPath: /config
         securityContext:
-          runAsNonRoot: true
           readOnlyRootFilesystem: true
       volumes:
       - name: dvo-config
@@ -70,10 +69,3 @@ spec:
       restartPolicy: Always
       serviceAccountName: deployment-validation-operator
       terminationGracePeriodSeconds: 30
-      resources:
-        requests:
-          memory: "200Mi"
-          cpu: "100m"
-        limits:
-          memory: "400Mi"
-          cpu: "200m"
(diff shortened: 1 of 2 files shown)


deploy/openshift/operator.yaml
```diff
@@ -20,12 +20,31 @@ spec:
...
+                matchExpressions:
+                - key: app
+                  operator: In
+                  values:
+                  - deployment-validation-operator
+              topologyKey: kubernetes.io/hostname
+            weight: 100
       containers:
-      - image: quay.io/deployment-validation-operator/dv-operator:latest
+      - image: quay.io/deployment-validation-operator/dv-operator:0.1.1
```
rrati, line 36: Not necessarily related to this topic, but we'll need some kind of documented release process/make target that will update this to latest release.  Maybe better to templatize this.

deploy/openshift/operator.yaml
```diff
@@ -40,13 +59,16 @@ spec:
         volumeMounts:
         - name: dvo-config
           mountPath: /config
+        securityContext:
+          runAsUser: 1000
```
rrati, line 63: We don't want to do a runAsUser.  This requires elevated privs on openshift.

rrati (changes requested):
Looks good.  Few minor comments

deploy/openshift/operator.yaml
bkez322: Hm, so you propose we just remove `runAsUser` from this and leave it at that? 

I'm okay with that, although DVO will then still complain about the `run_as_non_root` validation on itself.

Do you think we should add `run_as_non_root` to the [incompatible checks list](https://github.com/app-sre/deployment-validation-operator/blob/47264bae4f9a1f9514f5f8423082e6c0637219ab/pkg/validations/validation_engine.go#L178-L184)?

rrati: No, run_as_non_root is fine for a kubernetes environment.  For openshift however, pods are run as non-root by default because they are run as a random UID.  Using runAsUser requires elevated privs either way, but will be problematic on openshift w/o any benefit.

rrati: For now let's not add the runAsUser.

bkez322: It has been removed.

rrati (approved):
/lgtm

PR #121, opened 2021-11-10
Refactoring initializing engine due to new custom check conflicts

npecka:
When creating new custom checks within the validations package, if validation engine gets initialized  more then once, GO will throw a panic error. This is work set to remedy this issue.

### DIFF
--- a/pkg/validations/base_test.go
+++ b/pkg/validations/base_test.go
@@ -177,7 +177,7 @@ func TestIncompatibleChecksAreDisabled(t *testing.T) {
 
 	err := intializeEngine()
 	if err != nil {
-		t.Errorf("Error getting prometheus metric: %v", err)
+		t.Errorf("Error initializing engine: %v", err)
 	}
 
 	badChecks := getIncompatibleChecks()
--- a/pkg/validations/base_test.go
+++ b/pkg/validations/base_test.go
@@ -96,7 +96,7 @@ func intializeEngine(t *testing.T, customCheck ...config.Check) {
 			engine.config.CustomChecks[0] = customCheck[0]
 			return
 		}
-		// Intialize engine
+		// Initialize engine
 		e, err := newEngine(newEngineConfigWithCustomCheck(customCheck[0]))
 		if err != nil {
 			t.Errorf("Error creating validation engine %v", err)
@@ -106,17 +106,17 @@ func intializeEngine(t *testing.T, customCheck ...config.Check) {
 		if initializeFlagAllChecks == 1 {
 			return
 		}
-		// Intialize engine
+		// Initialize engine
 		e, err := newEngine(newEngineConfigWithAllChecks())
 		if err != nil {
 			t.Errorf("Error creating validation engine %v", err)
 		}
 		engine = e
-		intializeFlagAllChecks = 1
+		initializeFlagAllChecks = 1
 	}
 
-	// Set intialize flag
-	intializeFlag = 1
+	// Set Initialize Flag
+	initializeFlag = 1
 }
 
 func TestRunValidationsIssueCorrection(t *testing.T) {
--- a/pkg/validations/base_test.go
+++ b/pkg/validations/base_test.go
@@ -54,26 +54,16 @@ func newCustomCheck() config.Check {
 	}
 }
 
-func newEngineConfigWithCustomCheck(customCheck ...config.Check) config.Config {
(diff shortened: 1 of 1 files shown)


pkg/validations/base_test.go
```diff
@@ -85,13 +87,36 @@ func createTestDeployment(replicas int32) (*appsv1.Deployment, error) {
...
 
+func intializeEngine(t *testing.T, customCheck ...config.Check) {
+
+	// Reset global prometheus registry to avoid testing conflicts
+	metrics.Registry = prometheus.NewRegistry()
+
+	// Check if custom check has been set
+	if len(customCheck) > 0 {
+		// Initialize engine with custom check
+		e, err := newEngine(newEngineConfigWithCustomCheck(customCheck[0]))
```
rrati, line 98: Would actually be better to refactor this to support multiple custom checks vs just always using the first one here.  Doing this makes the function signature misleading.

pkg/validations/base_test.go
```diff
@@ -85,13 +87,36 @@ func createTestDeployment(replicas int32) (*appsv1.Deployment, error) {
...
+		if err != nil {
+			t.Errorf("Error creating validation engine %v", err)
```
rrati, line 100: Make the error message here mention using a custom check

rrati (changes requested):

pkg/validations/base_test.go
npecka: @rrati Can you double check the new changes? I put a loop in to iterate through the rest of the list once the engine is created if there are more custom checks.

rrati: There's no need to do this.  The custom checks in the config is an array, and what is passed in to this function is an array.  I would just modify newEngineWithCustomCheck to accept an array of check configs.

npecka: @rrati Changed it so that it fills in the newEngineWithCustomCheck function. Can you check it and see if that's what you were looking for?

rrati: This is still only using 1 check.  I was looking more for something like:

newEngineConfigWithCustomCheck(customCheck []config.Check) {
...
customConfig := config.Config{
		CustomChecks: customCheck,
...
...

npecka: @rrati Ah yes, I was over complicating it. The latest change should be using the array properly now.

pkg/validations/base_test.go
```diff
@@ -85,13 +87,34 @@ func createTestDeployment(replicas int32) (*appsv1.Deployment, error) {
...
+
+	// Reset global prometheus registry to avoid testing conflicts
+	metrics.Registry = prometheus.NewRegistry()
+
+	// Check if custom check has been set
+	if len(customCheck) > 0 {
+		// Initialize engine with custom check
+		e, err := newEngine(newEngineConfigWithCustomCheck(customCheck))
+		if err != nil {
+			t.Errorf("Error creating validation engine with custom checks %v", err)
```
rrati, line 100: Better to return an error here and let the caller deal with it.  That would also remove the need to pass in the testing.T.

pkg/validations/base_test.go
```diff
@@ -147,9 +174,10 @@ func TestRunValidationsIssueCorrection(t *testing.T) {
 }
 
 func TestIncompatibleChecksAreDisabled(t *testing.T) {
-	e, err := newEngine(newEngineConfigWithAllChecks())
+
+	err := intializeEngine()
 	if err != nil {
-		t.Errorf("Error creating validation engine %v", err)
+		return
```
rrati, line 180: Why did you change this?  If the engine fails to initialize, it should be printing a error like it was before

npecka: This was a mistake. Should of reflected similarly to the function above it. Fixed with latest change.

pkg/validations/base_test.go
```diff
@@ -147,9 +174,10 @@ func TestRunValidationsIssueCorrection(t *testing.T) {
 }
 
 func TestIncompatibleChecksAreDisabled(t *testing.T) {
-	e, err := newEngine(newEngineConfigWithAllChecks())
+
+	err := intializeEngine()
 	if err != nil {
-		t.Errorf("Error creating validation engine %v", err)
+		t.Errorf("Error getting prometheus metric: %v", err)
```
rrati, line 180: This should be the same error text as before?

rrati (approved):
/lgtm

PR #122, opened 2021-11-30
Make network policy yaml generic and update docs for usage

rrati:


README.md
```diff
@@ -59,6 +65,14 @@ do
 done
 ```
 
+## Allow scraping from outside DVO namespace
+
+The metrics generated by DVO can be scraped by anything that understand prometheus metrics.  A network policy may be needed to allow the DVO metrics to be collected from a service running in a namespace other than the one where DVO is deployed.  For example, if a service in `some-namespace` wants to scrape the metrics from DVO then a network policy would need to be created like this:
```
bkez322, line 70:
```suggestion
The metrics generated by DVO can be scraped by anything that understands prometheus metrics.  A network policy may be needed to allow the DVO metrics to be collected from a service running in a namespace other than the one where DVO is deployed.  For example, if a service in `some-namespace` wants to scrape the metrics from DVO then a network policy would need to be created like this:
```

bkez322 (approved):
Found a small typo.

Otherwise this makes sense to combine the NetworkPolicy objects. LGTM.

PR #126, opened 2021-12-06
expose minimal prometheus metrics (DVO-23)

bkez322:
Currently, DVO's `/metrics` endpoint exposes a great number of metrics, including those of the default prometheus go-client collectors, as well as metrics managed by the controller-runtime package.

This PR disables the `/metrics` endpoint managed by controller-runtime, and replaces it with a default and more minimal promhttp handler. In addition to the failing-validation metrics, this will cause DVO to expose only a few dozen non-validation metrics (e.g. goroutines, cpu, memory, http requests handled, etc) about DVO itself (as opposed to thousands of additional controller-runtime-managed metrics).

Prometheus-setup code is added to its own new package. This package manages the global Prometheus registry(which used to managed by controller-runtime).

Testing this on the app-sre-stage-01 cluster, this PR brings us down from 5k metrics to 2k metrics (99% of which begin with the `deployment_validation_operator_` prefix).

I have chosen to have this PR still expose a few dozen non-validation metrics (e.g. goroutines, cpu, memory, http requests handled, etc). I believe these metrics are invaluable for monitoring and troubleshooting DVO itself. Alternatively, if we truly want DVO to drop **_all_** non-validation metrics, that's easy to do, but this would be at the cost of operability of DVO imo.

Consideration for the future:
* Do we want there to be an optional configuration property to enable/disable the controller-runtime metrics on DVO? I'm thinking probably not, but this would be pretty easy to implement.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -34,9 +34,9 @@ import (
 
 // Change below variables to serve metrics on different host or port.
 var (
-	metricsPort       	int32 	= 8383
-	metricsPath			string	= "metrics"
-	defaultConfigFile   	    = "config/deployment-validation-operator-config.yaml"
+	metricsPort       int32  = 8383
+	metricsPath       string = "metrics"
+	defaultConfigFile        = "config/deployment-validation-operator-config.yaml"
 )
 var log = logf.Log.WithName("DeploymentValidation")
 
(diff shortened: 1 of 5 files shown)


rrati (approved):
/lgtm

PR #141, opened 2022-01-06
Add development environment

TGPSKI:
This MR proposes an all-inclusive development environment for DVO, based on k3d, tilt, and a few helper scripts.

This MR contains: 

* kubernetes manifests for DVO (`./deploy/kubernetes/*`)
* development environment (`./develop/*`)
    * install script
    * boostrap script
    * Tiltfile
    * customized dockerfile and manifests for use with Tilt

https://user-images.githubusercontent.com/12532803/148307889-987a6df6-b637-447e-9116-cbd8b554a3e9.mp4

### DIFF
--- a/deploy/kubernetes/operator.yaml
+++ b/deploy/kubernetes/operator.yaml
@@ -1,0 +1,58 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: deployment-validation-operator
+  name: deployment-validation-operator
+  namespace: deployment-validation-operator
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      app: deployment-validation-operator
+  strategy:
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        name: deployment-validation-operator
+        app: deployment-validation-operator
+    spec:
+      containers:
+      - image: quay.io/deployment-validation-operator/dv-operator:latest
+        imagePullPolicy: Always
+        name: deployment-validation-operator
+        args:
+        - --config /config/deployment-validation-operator-config.yaml
+        env:
+        - name: WATCH_NAMESPACE
+          value: ""
+        - name: OPERATOR_NAME
+          value: "deployment-validation-operator"
+        - name: NAMESPACE_IGNORE_PATTERN
+          value: "openshift.*|kube-.+"
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        volumeMounts:
+        - name: dvo-config
+          mountPath: /config
+        resources:
+          requests:
+            memory: "200Mi"
+            cpu: "100m"
+          limits:
+            memory: "400Mi"
+            cpu: "200m"
+      volumes:
+      - name: dvo-config
+        configMap:
+          optional: true
+          name: deployment-validation-operator-config
+      restartPolicy: Always
+      serviceAccount: deployment-validation-operator
+      terminationGracePeriodSeconds: 30
--- a/develop/README.md
+++ b/develop/README.md
@@ -1,0 +1,69 @@
+# Local Development Environment
+
+The scripts and manifests in this folder assist in creating a development environment for rapid iteration on `deployment-validation-operator`. Scripts for installing and bootstrapping the environment are included.
+
+Features:
+
+- Multi-node Kubernetes cluster in Docker with [K3D](https://k3d.io/)
+- Local image registry server
+- On golang / operator code changes, trigger automated container builds + Kubernetes workload updates with [Tilt](https://tilt.dev/)
+
+## Usage
+
+### Install & Configure
+
+1. Run install
+
+`$ bash install.sh`
+
+2. Edit `./develop/registries.yaml`:
+
+```
+configs:
+  quay.io:
+    auth:
+      username: ""
+      password: ""
+```
+
+3. Capture system-specific build command
+
+- Run `make` in deployment-validation-operator root directory
+- Capture build command (last command seen after running `make`) and save to `devel/Dockerfile`
+
+```
+$ cd deployment-validation-operator && make
+...
+...
+# Force GOOS=linux as we may want to build containers in other *nix-like systems (ie darwin).
+# This is temporary until a better container build method is developed
+GOOS=linux GOARCH=amd64 CGO_ENABLED=0 GOFLAGS=-mod=vendor GOOS=linux go build -gcflags="all=-trimpath=" -asmflags="all=-trimpath=" -o build/_output/bin/deployment-validation-operator ./cmd/manager
+
+$ vim devel/Dockerfile
+
+# diff
+- RUN <make command output here>
++ RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 GOFLAGS=-mod=vendor GOOS=linux go build -gcflags="all=-trimpath=" -asmflags="all=-trimpath=" -o build/_output/bin/deployment-validation-operator ./cmd/manager
+```
+
+### Bootstrap
+
+`$ bash bootstrap.sh`
+
+### Develop
+
+`$ tilt up`
+
+### Teardown
+
+`$ k3d cluster delete dvo-local`
+
+## Tooling
+
+### K3D
+
+[K3D](https://k3d.io/)
+
+### Tilt
+
+[Tilt](https://tilt.dev/)
--- a/develop/bootstrap.sh
+++ b/develop/bootstrap.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+
+CLUSTER_NAME="dvo-local"
+PROMETHEUS_NAMESPACE="prom"
+GRAFANA_NAMESPACE="graf"
+DEVELOPMENT_NAMESPACE="check"
+DVO_NAMESPACE="deployment-validation-operator"
+
+INSTALL_PROMETHEUS="true"
+INSTALL_PROMETHEUS_OPERATOR="false"
+INSTALL_GRAFANA="false"
+
+k3d cluster list | grep ${CLUSTER_NAME} &> /dev/null
+SUCCESS=$?
+if [ $SUCCESS -eq 1 ]
+then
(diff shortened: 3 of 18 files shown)


rrati:
Can we re-org things so that we don't duplicate deployment yamls?  It'll make keeping changes up to date tricky.

develop/bootstrap.sh
```diff
@@ -0,0 +1,81 @@
...
+        echo "Error selecting k3d kubernetes context"
+        exit 1
+    fi
+fi
+
+if [ "$INSTALL_PROMETHEUS" == "true" ]
+then
+    echo "Installing prometheus"
+    echo ""
+    helm install -n $PROMETHEUS_NAMESPACE --create-namespace --set server.service.type=NodePort --set server.service.nodePort=30080 prom prometheus-community/prometheus &> /dev/null
```
rrati, line 44: Can we do these actions without growing a dep on helm?  We'll already have deps on oc, kubectl, tilt, k3d at the very least.  I'd like the dep list to be as small as possible.

rrati: NM.  Helm is installed and managed later in the script.

rrati:
Also recommend adding a "Contributing", "Getting Started", or similar section to the main README and reference the one in develop

TGPSKI:
I'm attempting to simplify the development environment and reduce duplication of files, as well as simplify the setup process.

I re-visited my attempts at using the stock dockerfile, but I continue to run into problems related to make.

```
  USER ${USER_UID}


     Building image
     Sending Docker build context: 87.79MB (648ms)
     copy /context / [done: 366ms]
     [stage-1 1/4] FROM registry.access.redhat.com/ubi8/ubi-minimal:latest@sha256:cf1c63e3247e4074ee3549a064b8798a1a2513ad57dd79c9edb979836355b469
     [builder 1/5] FROM quay.io/app-sre/boilerplate:image-v1.0.0
     [builder 2/5] RUN mkdir -p /workdir [cached]
     [builder 3/5] COPY . /workdir [cached]
     [builder 4/5] WORKDIR /workdir [cached]
     [builder 5/5] RUN make
       → Makefile:23: warning: overriding recipe for target `build-push'
       → boilerplate/openshift/golang-osd-operator/standard.mk:234: warning: ignoring old recipe for target `build-push'
       → boilerplate/openshift/golang-osd-operator/ensure.sh golangci-lint
       → fatal: not a git repository (or any of the parent directories): .git
       → fatal: not a git repository (or any of the parent directories): .git
       → fatal: not a git repository (or any of the parent directories): .git
       → + set -eo pipefail
       → ++ git rev-parse --show-toplevel
       → fatal: not a git repository (or any of the parent directories): .git
       → + REPO_ROOT=
       → make: *** [go-check] Error 128
     
     ERROR IN: [builder 5/5] RUN make

Build Failed: ImageBuild: executor failed running [/bin/sh -c make]: exit code: 2
```

TGPSKI:
@rrati @bkez322 I found a workaround in Tilt to remove the dependency on a second Dockerfile. This removes the need to capture a system-specific build command and maintain a development Dockerfile. 

The downside is that `RUN make` is not cachable at the docker build level, and this adds +5 minutes to build and deploy a new code change. If you are quickly making changes and want to see almost instant builds, then only building and deploying using the method described originally in the PR is much faster.

My opinion: it feels like it takes forever to build after I made the change. This is caused by additional steps in the make build operation which runs linting, tests, downloads submodules, etc. 

Tradeoffs:

Original PR Approach
-------
* Duplicate Dockerfile to maintain
* Individual system-specific setup to enable development environment
* Much faster builds (not running submodule update, lint, and test before build)

Updated Approach
----------
* No duplicate Dockerfile
* No additional setup for Dockerfile (there still is setup required for registry credentials for quay.io)
* Much slower builds

rrati:
@TGPSKI If I understand correctly, could we just add a new target in the Makefile that reduces the amount of steps needed for this to work?  Like d/l modules shouldn't happen because it SHOULD be using the vendored modules.  Unit tests seems like it should be the same time this way vs the other method, but if other method doesn't run unit tests and speed is important vs correctness then maybe an ENV to set to enable/disable running the tests?

Basically, is there anything we can do in Makefile to cut down on the build time for this approach?

develop/install.sh
```diff
@@ -0,0 +1,38 @@
...
+if ! command -v helm &> /dev/null
+then
+    echo "* installing helm"
+    curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+fi
+echo "* helm installed"
+
+if ! command -v tilt &> /dev/null
+then
+    ehco "* installing tilt"
```
rrati, line 29: typo 'ehco' -> 'echo'

develop/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Usage
+
+### Install & Configure
+
+1. Run install
+
+`$ bash install.sh`
+
+2. Edit `./develop/registries.yaml`:
```
rrati, line 19: What is meant by this step?  Edit so username/password are empty strings or set them to values?

develop/README.md
```diff
@@ -0,0 +1,69 @@
...
+$ vim devel/Dockerfile
+
+# diff
+- RUN <make command output here>
++ RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 GOFLAGS=-mod=vendor GOOS=linux go build -gcflags="all=-trimpath=" -asmflags="all=-trimpath=" -o build/_output/bin/deployment-validation-operator ./cmd/manager
+```
+
+### Bootstrap
+
+`$ bash bootstrap.sh`
```
rrati, line 51: $ cd develop
$ bash bootstrap.sh

Alternately edit bootstrap.sh to be able to run from root of repo or develop dir.

rrati:
Tested and verified working on OSX

develop/README.md
```diff
@@ -0,0 +1,69 @@
...
+- Local image registry server
+- On golang / operator code changes, trigger automated container builds + Kubernetes workload updates with [Tilt](https://tilt.dev/)
+
+## Usage
+
+### Install & Configure
+
+1. Run install
+
+`$ bash install.sh`
```
bkez322, line 17: should maybe say:

Run install from the deployment-validation-operator root directory:
`$ bash develop/install.sh`

develop/README.md
bkez322: I didn't get a mac os security warning, but maybe it's something I accepted a long time ago and don't remember.

bkez322:
Works on my Mac OS as well. Awesome work on this/

develop/README.md
```diff
@@ -0,0 +1,69 @@
...
++ RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 GOFLAGS=-mod=vendor GOOS=linux go build -gcflags="all=-trimpath=" -asmflags="all=-trimpath=" -o build/_output/bin/deployment-validation-operator ./cmd/manager
+```
+
+### Bootstrap
+
+`$ bash bootstrap.sh`
+
+### Develop
+
+`$ tilt up`
```
bkez322, line 55: I'm a dummy and it took me a few moments to realize `$tilt up` didn't _immediately_ deploy the app, and so I was confused when doing `kubectl get pods -A` immediately after didn't show me the expected DVO pod. Now I know it takes tilt ~45 seconds (on my laptop) to build the image and deploy it.

For future dummies like me, it may be worth adding under `$tilt up`:
>Check the tilt web UI (press space) to see progress on tilt building and deploying the DVO image.

bkez322:
>The downside is that RUN make is not cachable at the docker build level, and this adds +5 minutes to build and deploy a new code change. If you are quickly making changes and want to see almost instant builds, then only building and deploying using the method described originally in the PR is much faster.

@TGPSKI 

I tested with this PR on this branch as-is. In my small bit of testing with very-small code-changes, the build time seemed to stay with 45-60 seconds. I'm not sure why the big difference between that and your 5 minutes.

How short were the build+deploy times for you before the dockerfile change?

TGPSKI:
@bkez322 the times you report are accurate.

Imagine this use case:

* You are debugging a corner case in DVO which involves kubernetes logic / return variables
* You set a logging statement to get some real-time data on interactions between kube-api-server and dvo
* You hit save, wait +5 minutes for the rebuild.
* You realize that you misspelled a variable name, so you get an error. Fix the variable name, and rebuild. +5
* The variable doesn't have all the info you need, so you add another logging statement. Rebuild, +5
* You find the bug, and then add code to fix it. Save + rebuild, +5 minutes.
* You finalize your change, rebuild. +5

Build time: 25 minutes. Real time, probably an 1-1.5 hours

Now, build with the fast option:

* You are debugging a corner case in DVO which involves kubernetes logic / return variables
* You set a logging statement to get some real-time data on interactions between kube-api-server and dvo
* You hit save, wait +1 minutes for the rebuild.
* You realize that you misspelled a variable name, so you get an error. Fix the variable name, and rebuild. +1
* The variable doesn't have all the info you need, so you add another logging statement. Rebuild, +1
* You find the bug, and then add code to fix it. Save + rebuild, +1 minutes.
* You finalize your change, rebuild. +1

Build time: 5 minutes. Real time, 30-45 minutes.

hlipsig:
I think all the previous issues were addressed, but I will need to test this or make sure Alex is ready to sign off on the merge.

deploy/kubernetes/operator.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: deployment-validation-operator
+  name: deployment-validation-operator
+  namespace: deployment-validation-operator
+spec:
+  replicas: 3
```
hlipsig, line 9: We know that more than 1 replica doesn't currently work, may be worthwhile to reduce to 1 until that's fixed.

bkez322: Good catch.

bkez322:
> I think all the previous issues were addressed, but I will need to test this or make sure Alex is ready to sign off on the merge.

FWIW I think Tyler has every intention of revisiting and wrapping up this merge request soon-ish.

tremes:
Hi @TGPSKI, what is the status of this PR please? I don't think any dev environment is needed and personally I wouldn't add any more manifests (because it's already confusing sometimes).

PR #149, opened 2022-05-26
add probes to deployment

maorfr:
DVO should implement best practices it itself recommends.

since DVO's main function is to provide metrics related to DVO violations, IMO it makes sense to use the metrics endpoint as an indicator for readiness and health.

### DIFF
--- a/deploy/openshift/operator.yaml
+++ b/deploy/openshift/operator.yaml
@@ -56,6 +56,9 @@ spec:
         readinessProbe:
           tcpSocket:
             port: 8383
+          initialDelaySeconds: 30
+          periodSeconds: 10
+          timeoutSeconds: 3
         env:
         - name: WATCH_NAMESPACE
           value: ""
--- a/deploy/openshift/operator.yaml
+++ b/deploy/openshift/operator.yaml
@@ -47,6 +47,17 @@ spec:
           limits:
             memory: "4Gi"
             cpu: "1"
+        livenessProbe:
+          httpGet:
+            path: /metrics
+            port: 8383
+          initialDelaySeconds: 30
+          periodSeconds: 10
+          timeoutSeconds: 3
+        readinessProbe:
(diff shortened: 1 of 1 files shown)


deploy/openshift/operator.yaml
```diff
@@ -47,6 +47,17 @@ spec:
           limits:
             memory: "4Gi"
             cpu: "1"
+        livenessProbe:
+          httpGet:
+            path: /metrics
```
rporres, line 52: in DVO this is a very data heavy endpoint which I don't think is wise to query every 10s in addition to what Prometheus itself is doing. If we don't have a proper `/health` endpoint, I guess we'd rather start with a tcp probe.

BumbleFeng:
lgtm

deploy/openshift/operator.yaml
```diff
@@ -47,6 +47,15 @@ spec:
...
             cpu: "1"
+        livenessProbe:
+          tcpSocket:
+            port: 8383
+          initialDelaySeconds: 30
+          periodSeconds: 10
+          timeoutSeconds: 3
+        readinessProbe:
+          tcpSocket:
+            port: 8383
```
rporres, line 58: since you're at it and you're targeting the same port, I would define the same option as in the liveness probe

```
          initialDelaySeconds: 5
          periodSeconds: 10
          timeoutSeconds: 3
```

maorfr: done: https://github.com/app-sre/deployment-validation-operator/pull/149/commits/0c527a7487dea37bbe9275242a3ab10a1aafa44a

maorfr: really done: https://github.com/app-sre/deployment-validation-operator/pull/149/commits/2e33b8feb03f379d739783adab0e4487c44fb32d

rporres (approved):

PR #157, opened 2022-09-01
Refactor to timer

thetechnick:
DVO is caching _a lot_ of different objects throughout a cluster,
leading to huge caches. As DVO does not need to check objects on every
single update and is only accessing each object state once, so the
allocated caches are rather underutilized.

This PR disables caches completely and replaces active watching with
a simple polling job.

Prerequisite for this PR is the refactoring of controller creation and GVK discovery to inject List GVK into each generic controller. Ideally using `RestMapper` instead iterating over a compile-time scheme.


nikhil-thomas:
development on this pr is continued in 
https://github.com/app-sre/deployment-validation-operator/pull/163

hence closing this pr

PR #158, opened 2022-09-15
Iniital olm artifacts template

pepedocs:
**What**
Created an initial olm artifacts template based on what I can infer from the olm artifacts template in addon-operator.

**Why**
App-interface saas file will need to refer to this template file in order to create a selectorsyncset in hive.

I'm new at DVO and the hive/olm/app-interface deployment model/processes so I may have naively created this template in that case please let me know. I partially tested dvo deployment on a local kind cluster with olm and used another hive kind cluster to test the SS.


jmelis (approved):
LGTM

PR #160, opened 2022-09-20
Added namespaces to ignore for platform

npecka:
Quick change to ensure OSD rollout of DVO accounts for all namespaces to ignore to ensure customer does not receive platform related recommendations


jmelis (approved):
LGTM

PR #161, opened 2022-09-21
Adding network object for insight operator and namespace pattern ignore

npecka:
Adding network policy to allow ingress from only the openshift insights operator. Also added namespace pattern exclusions to not have DVO alert on platform level stuff to customer.


jmelis (approved):
LGTM

PR #163, opened 2022-09-30
Reduce operator memory consumption

nikhil-thomas:
Reduce memory consumption

Run the operator logic in 1 reconciler (thread) instead
of creating 1 reconciler (thread) per api resource

- DVO is caching a lot of different objects throughout a cluster,
leading to huge caches. As DVO does not need to check objects on every
single update and is only accessing each object state once, so the
allocated caches are rather underutilized.

- This PR disables caches completely and replaces active watching with
a simple polling job.

Add a mechanism to shortlist namespaces where validations should be run
for namespaced resources. This list is used during the list resources logic
so that resource instances from ignored namespaces are not fetched from api server.

Add pagination (limit-continue) in the list calls. The usage of pagination
ensures that the memory consumption by this operator always stays under
a desired threshold irrespective of the number of resource instances
of any kubernetes resource

Replace client-go scheme.Scheme based static creation (compile time decision) of the list of api resources being watched (polled) with discovery client based mechanism. The new mechanism creates the list of api resources that must be validated during runtime.


nikhil-thomas:
closing this as the newer pr https://github.com/app-sre/deployment-validation-operator/pull/165 has this work included in it.

PR #164, opened 2022-09-30
Adding skipRange for OSD deployment

npecka:
Adding skipRange instead of replaces for OSD CSV


jmelis:
Closing this one. In the end we've decided to use the 'SKIP' process that is already supported by boilerplate.

PR #166, opened 2022-10-03
Restrict rollout of SSS

cblecker:
This adds in additional restrictions to the rollout of DVO:
- Only roll out to clusters 4.11+
- Only roll out to clusters in the staging and integration environments


cblecker:
/assign @jmelis 
cc @jharrington22

jharrington22 (commented):
/lgtm

We can and production and a org selector when you are ready to deploy @jmelis

PR #167, opened 2022-10-04
Deploy only to target org

jmelis:


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -43,6 +47,9 @@ objects:
           - key: api.openshift.com/environment
             operator: In
             values: ["staging", "integration"]
+          - key: api.openshift.com/legal-entity-id
+            operator: In
+            values: ["${TARGET_ORG}"]
```
jharrington22, line 52: This would cause DVO to be remove from clusters in stage, you could change the `matchLabelApplyMode` to `OR`

This would be on the same level as `matchExpressions` 

```
matchLabelsApplyMode: "OR"
```

So for stage and integration it would be applied everywhere and for production just on clusters with TARGET_ORG

We can then remove the `matchLabelsApplyMode` and TARGET_ORG expression when we are ready.

cblecker: `matchLabelsApplyMode` is an MCC unique thing -- it's not a part of the [upstream spec](https://github.com/openshift/hive/blob/0ca68fed9e3b135e823d349c2da7dc8302dc98bb/apis/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1172-L1191).

I'd suggest 
```suggestion
            values: "${{TARGET_ORG}}"
```
similar to https://github.com/openshift/managed-cluster-config/blob/master/deploy/cloud-ingress-operator-configuration/routerreplicas-osd-8028/config.yaml, so that we can add multiple org IDs in there to cover any staging env orgs we want to test with, and the appsre prod org.

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -43,6 +47,9 @@ objects:
           - key: api.openshift.com/environment
             operator: In
             values: ["staging", "integration"]
+          - key: api.openshift.com/legal-entity-id
+            operator: In
+            values: "${{TARGET_ORG}}"
```
jharrington22, line 52: You need to remove the quotes here see [this](https://github.com/openshift/pagerduty-operator/blob/master/hack/olm-registry/olm-artifacts-template.yaml#L91) example.

jharrington22 (changes requested):

jharrington22 (approved):
/lgtm

hack/olm-registry/olm-artifacts-template.yaml
jmelis: done!

PR #169, opened 2022-10-05
feat: adding namespace_uid and name_uid to metrics

rluders:
TBA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ build/_output/*
 .operator-sdk/*
 .grpcurl/*
 .opm/*
+
+# goland
+.idea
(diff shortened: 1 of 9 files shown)


nikhil-thomas:
early feedback: we are listing namespaces https://github.com/app-sre/deployment-validation-operator/blob/daa2651fb136f5e9c9b28f2b457e782e120d5a24/pkg/controller/watchnamespaces.go#L14 already. This happens once every loop.

Why don't we modify the `watchNamespaceCache` struct to something like
```
type struct namespace {
  name, uuid string
}

type watchNamespacesCache struct {
	namespaces    *[]namespace
	ignorePattern *regexp.Regexp
}
```

then we can look up the namespaceuuid with a namespace name.

Alternatively, we can consider doing the following:
```
type watchNamespacesCache struct {
	namespaces    map[string]string // name -> uuid
	ignorePattern *regexp.Regexp
}
```

PR #184, opened 2022-10-11
Replace error return to manager with a log statement

nikhil-thomas:
This is a temporary fix for https://issues.redhat.com/browse/DVO-72)
 to avoid pod crash and restart.

The issue is caused due to version skew in server kubernetes version and client-go version used in the operator code. A proper fix will be pushed in a different patch.

This patch will make the "reconciler" loop locally loop and continue with other resources without resulting in a pod crash.


Ajpantuso (approved):
/lgtm

PR #185, opened 2022-10-12
fix: ensure a single scheme is used throughout codebase

Ajpantuso:
### Summary

- Propagates a single scheme up through `main`.
- Adds error wrapping where missing along the "core" code path.


pkg/controller/generic_reconciler.go
```diff
@@ -7,14 +7,13 @@ import (
 	"os"
 	"strconv"
 
-	osappsscheme "github.com/openshift/client-go/apps/clientset/versioned/scheme"
+	"go.uber.org/multierr"
```
nikhil-thomas, line 10: ❤️

PR #187, opened 2022-10-12
Improve apiResource listing mechansim for deciding resources to be validated

nikhil-thomas:
Add a check to ensure that an APIResource (gvk) supported by the server is registered in the client-go scheme

Add a mechanism to make sure that the gvk of the APIResource is chosen based on what is supported by client scheme when server supports more than one version.

eg: server supports `batch/v1/CronJobs` and `batch/v1beta1/CronJob` but client knows only about 
`batch/v1beta1/CronJob`. In this case `batch/v1beta1/CronJob` will be used in queries.

### DIFF
--- a/pkg/controller/generic_reconciler.go
+++ b/pkg/controller/generic_reconciler.go
@@ -113,9 +113,16 @@ func (gr *GenericReconciler) Start(ctx context.Context) error {
 }
 
 func (gr *GenericReconciler) reconcileEverything(ctx context.Context) error {
-	apiResources, err := reconcileResourceList(gr.discovery)
+	var log = logf.Log.WithName("reconcileEverything")
+	apiResources, err := reconcileResourceList(gr.discovery, gr.client.Scheme())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


nikhil-thomas:
/hold
minor improvement pending

Ajpantuso (commented):
Logically I don't see any issues, but wrote a quick refactor to demonstrate ways to add clarity to the nested loops. See https://github.com/app-sre/deployment-validation-operator/commit/221e767a29f099378fedc16217e458d4dba6a3c7 for some suggestions.

nikhil-thomas:
https://github.com/app-sre/deployment-validation-operator/commit/221e767a29f099378fedc16217e458d4dba6a3c7 definitely adds more clarity. The long nested loop did seem ugly. Thanks for the suggestion. I shall make necessary changes 🧑‍💻 👍

nikhil-thomas:
@Ajpantuso i have refactored the patch as per your suggestion.
ptal 🧑‍💻

Ajpantuso (approved):
/lgtm

PR #190, opened 2022-10-13
Restrict deploying DVO to clusters that are 4.11.10+

cblecker:
This ensures that clusters need to be 4.11.10+ to get DVO.


cblecker:
Note that we shouldn't merge this until 4.11.10 is available, as even test clusters won't be able to upgrade to it.

npecka (approved):

PR #193, opened 2022-10-14
Add a workqueue to the reconciler

nikhil-thomas:
Replace "continuously loop through" all resource types with a "workquue" based mechanism

This patch decouples 2 parts of our reconciler:

1. deciding the resource types to be validated and setting them up for processing ("work manager" or events)
2. fetching resource instances of each resource types and running validations ("worker" or event handling)

This patch requeues resource types which failed (returned error) on fetching resource instances or while running validations

TODO: making the "validaiton reslult cache" and "handle deletion" mechanisms thread safe


pkg/controller/reconcileobjects.go
```diff
@@ -16,6 +16,33 @@ type resourceSet struct {
 	apiResources map[schema.GroupKind]metav1.APIResource
 }
 
+type hasableAPIResource struct {
```
thetechnick, line 19:
```suggestion
type hashableAPIResource struct {
```

pkg/controller/generic_reconciler.go
```diff
@@ -257,16 +283,21 @@ func (gr *GenericReconciler) lookUpType(obj *unstructured.Unstructured) (runtime
...
 		}
-		validations.DeleteMetrics(k.namespace, k.name, k.kind)
-		gr.objectValidationCache.removeKey(k)
-
+		nsUID := gr.watchNamespaces.getNamespaceUID(k.namespace)
+		validations.DeleteMetrics(k.namespace, nsUID, k.name, k.uid, gvk.Kind)
+		objectHistory.removeKey(k)
 	}
-	gr.currentObjects.drain()
+	currentObjects = nil
```
thetechnick, line 300: What is this supposed to do?
This just set's the local scoped variable to `nil`

pkg/controller/generic_reconciler.go
```diff
@@ -93,27 +96,37 @@ func (gr *GenericReconciler) AddToManager(mgr manager.Manager) error {
 // Start validating the given object kind every interval.
 func (gr *GenericReconciler) Start(ctx context.Context) error {
 	var log = logf.Log.WithName("controller")
+	ticker := time.Tick(60 * time.Second)
+	defer gr.workqueue.ShutDown()
+
+	workerCount := 3
```
thetechnick, line 102: Running multiple workers only makes sense, if you increase QpS settings, otherwise having multiple workers waiting for QpS is not helping throughput :)

pkg/controller/generic_reconciler.go
nikhil-thomas: oops

pkg/controller/reconcileobjects.go
nikhil-thomas: fixed

pkg/controller/generic_reconciler.go
nikhil-thomas: i have added environment variables to configure:
- qps 
- enqueue interval
- number of workers

Therefore these values could be tunned during deployment.

nikhil-thomas:
/hold

need to cache promlabels as well to make sure `handleDeletion` is thread safe.

PR #206, opened 2022-10-24
fix: code clean up

rluders:
Folks, this is just a clean-up based on some shallow analysis. Up to you to decide if it makes sense or not.


rluders:
Not relevant anymore.

PR #216, opened 2022-11-11
Add main package unit tests and some refactor

ncaak:
## Refactor PR

Little change to move a struct and its functionality from main package (but different file to main.go) to an existing configuration package.
Adding some unit tests taking advantage on new file distribution.

### actions

* Moved configuration struct (options) to **config** package
* Fixed accesses for the struct and little tweak
* Added few unit tests for **main** package functions
* Added unit tests for **config** package
* Minor change on gitignore for vscode devs


cmd/manager/main.go
```diff
@@ -12,7 +12,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
 
-	dv_config "github.com/app-sre/deployment-validation-operator/config"
+	dvConfig "github.com/app-sre/deployment-validation-operator/config"
```
Ajpantuso, line 15:
```suggestion
	dvconfig "github.com/app-sre/deployment-validation-operator/config"
```

The Go preference for package names is all lowercase rather than camel case or snake case.

cmd/manager/main_test.go
```diff
@@ -0,0 +1,42 @@
+package main
+
```
Ajpantuso, line 2: Rather than adding tests to main I think it would make sense to use the integration suite https://github.com/app-sre/deployment-validation-operator/tree/master/integration. This should reduce coupling between tests and implementation.

config/options.go
```diff
@@ -1,4 +1,4 @@
-package main
+package config
```
Ajpantuso, line 1: Instead of moving this to `config` and adding to the public API surface of this module I think `internal/cmd` or a similar internal package would be a better place for the `Options` struct.

Ajpantuso (changes requested):
A few recommendations, but overall these should be useful for preventing regressions, thanks!

Ajpantuso (changes requested):

internal/options/options.go
```diff
@@ -9,33 +9,33 @@ import (
...
 )
 
-type options struct {
+type Options struct {
 	MetricsPort    int32
 	MetricsPath    string
 	ProbeAddr      string
 	ConfigFile     string
-	watchNamespace *string
+	WatchNamespace *string
```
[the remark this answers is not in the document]
ncaak: Yes, this was a typo after dropping main package tests. Fixing this and pushing the change.

Ajpantuso (approved):
/lgtm

PR #218, opened 2022-11-25
Remove outdated instructions

ncaak:
### Documentation PR

After Slack conversation, New Release doc was outdated. Adding some rewording and removing outdated entry.


npecka (changes requested):
These changes need to pass the failed checks first. Assure they are passing the lint checks before pushing the PR. If changes are pushed while these checks fail the build pipeline will become unstable and not feed changes.

npecka (approved):

PR #221, opened 2022-12-13
Controller unit tests

ncaak:
#### summary
Refactored Controller package : watchnamespaces functionality
* added some documentation to functions
* refactored getWatchNamespaces function, no changes on signature
* added unit tests

#### misc
* fixed Prometheus package unit tests wrong documentation

#### tracking
https://issues.redhat.com/browse/CCXDEV-9722


pkg/controller/watchnamespaces.go
```diff
@@ -30,10 +34,24 @@ func newWatchNamespacesCache() *watchNamespacesCache {
 	}
 }
 
+// getFormattedNamespaces returns a list of namespaces filtering through given list
+// and formatting them as namespace structs
+func getFormattedNamespaces(list corev1.NamespaceList, ignorePattern *regexp.Regexp) (fns []namespace) {
+	for _, ns := range list.Items {
+		name := ns.GetName()
+		if ignorePattern == nil || !ignorePattern.Match([]byte(name)) {
+			fns = append(fns, namespace{uid: string(ns.GetUID()), name: name})
```
tremes, line 43: Not much related to this update, but perhaps we could have the `uid` attribute as `types.UID` instead of string.

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,102 @@
+package controller
```
tremes, line 1: This test seems to test functions from `watchnamespaces.go` so I  am wondering whether it would be better to rename this file to `watchnamespaces_test.go` (instead of `controller_test.go`). WDYT?

pkg/controller/watchnamespaces.go
nikhil-thomas: i have no strong opinion here. I think the uid is mainly used as a value for a prometheus label. So `string` should be good enough. 🧑‍💻

nikhil-thomas:
/hold

feel free to remove this hold once the file is renamed.

pkg/controller/watchnamespaces.go
ncaak: I added a [note](https://github.com/app-sre/deployment-validation-operator/pull/221/files#diff-442af8c760ab7d47870e117c7b97eead9959b6f3bea8961f180efb7bb32e2c80R13) for future change. I didn't want to change any signature on this PR.

Ajpantuso (approved):
/lgtm

PR #222, opened 2022-12-19
Minor refactor on Controller package : validationscache

ncaak:
### summary
Refactored Controller package : validationscache functionality

* added some documentation to functions
* refactored **objectAlreadyValidated** function to reduce cognitive complexity
* added unit tests for **objectAlreadyValidated** and **store** methods

### tracking
https://issues.redhat.com/browse/CCXDEV-9722

### DIFF
--- a/pkg/controller/validationscache.go
+++ b/pkg/controller/validationscache.go
@@ -34,7 +34,8 @@ type validationResource struct {
 	outcome validations.ValidationOutcome
 }
 
-// newValidationResource returns a new empty instance of validationResource struct
+// newValidationResource returns a 'validationResource' populated
+// with the given 'resourceVersion', 'uid', and 'ValidationOutcome'.
 func newValidationResource(
 	rscVer resourceVersion,
 	uid string,
--- a/pkg/controller/validationscache.go
+++ b/pkg/controller/validationscache.go
@@ -101,7 +101,12 @@ func (vc *validationCache) retrieve(obj client.Object) (validationResource, bool
 	return val, exists
 }
 
-// objectAlreadyValidated returns if the given object has been validated
+// objectAlreadyValidated returns 'true' if the given 'Object'
+// has a cached 'ValidationOutcome' with the same 'ResourceVersion'
+// (Kubernetes representation of iteration count for a persisted resource).
+// If the 'ResourceVersion' of an existing 'Object' is stale the cached
+// 'ValidationOutcome' is removed and 'false' is returned. In all other
+// cases 'false' is returned.
 func (vc *validationCache) objectAlreadyValidated(obj client.Object) bool {
 	validationOutcome, exists := vc.retrieve(obj)
 	if exists {
--- a/pkg/controller/validationscache.go
+++ b/pkg/controller/validationscache.go
@@ -15,8 +15,8 @@ func newResourceversionVal(str string) resourceVersion {
 	return resourceVersion(str)
 }
 
-// newValidationKey returns an instance of validationKey struct
-// populated with data extrated from the given object
+// newValidationKey returns a unique identifier for the given
+// object suitable for hashing.
 func newValidationKey(obj client.Object) validationKey {
 	gvk := obj.GetObjectKind().GroupVersionKind()
 	return validationKey{
--- a/pkg/controller/validationscache.go
+++ b/pkg/controller/validationscache.go
@@ -79,8 +79,9 @@ func (vc *validationCache) drain() {
 	*vc = validationCache{}
 }
 
-// remove deletes a key, and its value, from the instance
-// it uses given object to search for the validationKey
+// remove uncaches the 'ValidationOutcome' for the
+// given object if it exists and performs a noop
+// if it does not.
 func (vc *validationCache) remove(obj client.Object) {
 	key := newValidationKey(obj)
 	vc.removeKey(key)
--- a/pkg/controller/validationscache.go
+++ b/pkg/controller/validationscache.go
@@ -73,8 +73,8 @@ func (vc *validationCache) store(obj client.Object, outcome validations.Validati
 	)
 }
 
-// drain overwrites the current instance with an empty one
-// all keys and values will be lost
+// drain frees the cache of any used resources
+// resulting in all cached 'ValidationOutcome's being lost.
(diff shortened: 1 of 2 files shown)


pkg/controller/validationscache.go
```diff
@@ -15,6 +15,8 @@ func newResourceversionVal(str string) resourceVersion {
 	return resourceVersion(str)
 }
 
+// newValidationKey returns an instance of validationKey struct
+// populated with data extrated from the given object
```
Ajpantuso, line 19:
```suggestion
// newValidationKey returns a unique identifier for the given
// object suitable for hashing.
```

pkg/controller/validationscache.go
```diff
@@ -32,6 +34,7 @@ type validationResource struct {
 	outcome validations.ValidationOutcome
 }
 
+// newValidationResource returns a new empty instance of validationResource struct
```
Ajpantuso, line 37:
```suggestion
// newValidationResource returns a 'validationResource' populated
// with the given 'resourceVersion', 'uid', and 'ValidationOutcome'.
```

pkg/controller/validationscache.go
```diff
@@ -46,15 +49,20 @@ func newValidationResource(
 
 type validationCache map[validationKey]validationResource
 
+// newValidationCache returns a new empty instance of validationCache struct
 func newValidationCache() *validationCache {
 	return &validationCache{}
 }
 
+// has returns a boolean if the given key exist in the instance
```
Ajpantuso, line 57:
```suggestion
// has returns 'true' if the given key exist in the instance; 'false' otherwise.
```

pkg/controller/validationscache.go
```diff
@@ -46,15 +49,20 @@ func newValidationResource(
...
 func (vc *validationCache) has(key validationKey) bool {
-	_, ok := (*vc)[key]
-	return ok
+	_, exists := (*vc)[key]
+	return exists
 }
 
+// store creates a key-value pair on the current instance
+// it uses given object to create a validationKey and a validationResource structs
+// constraint: if a key already exists, it will overwrite the value
```
Ajpantuso, line 65:
```suggestion
// store caches a 'ValidationOutcome' for the given 'Object'.
// constraint: cached outcomes will be updated in-place for a given object and
// consecutive updates will not preserve previous state.
```

Better not to reveal internal details to consumers of the API through documentation.

pkg/controller/validationscache.go
```diff
@@ -64,35 +72,45 @@ func (vc *validationCache) store(obj client.Object, outcome validations.Validati
 	)
 }
 
+// drain overwrites the current instance with an empty one
+// all keys and values will be lost
```
Ajpantuso, line 76:
```suggestion
// drain frees the cache of any used resources
// resulting in all cached 'ValidationOutcome's being lost.
```

pkg/controller/validationscache.go
```diff
@@ -64,35 +72,45 @@ func (vc *validationCache) store(obj client.Object, outcome validations.Validati
...
 func (vc *validationCache) drain() {
 	*vc = validationCache{}
 }
 
+// remove deletes a key, and its value, from the instance
+// it uses given object to search for the validationKey
```
Ajpantuso, line 82:
```suggestion
// remove uncaches the 'ValidationOutcome' for the
// given object if it exists and performs a noop
// if it does not.
```

pkg/controller/validationscache.go
```diff
@@ -64,35 +72,45 @@ func (vc *validationCache) store(obj client.Object, outcome validations.Validati
...
 }
 
+// removeKey deletes a key, and its value, from the instance
 func (vc *validationCache) removeKey(key validationKey) {
 	delete(*vc, key)
 }
 
+// retrieve returns the value, if exists, within a key
+// it uses given object to search for the validationKey
+// it returns a second parameter to check if the key existed in the instance
```
Ajpantuso, line 95:
```suggestion
// retrieve returns a tuple of 'validationResource' (if present)
// and 'ok' which returns 'true' if a 'validationResource' exists
// for the given 'Object' and 'false' otherwise.
```

pkg/controller/validationscache.go
```diff
@@ -64,35 +72,45 @@ func (vc *validationCache) store(obj client.Object, outcome validations.Validati
...
 func (vc *validationCache) retrieve(obj client.Object) (validationResource, bool) {
 	key := newValidationKey(obj)
-	val, ok := (*vc)[key]
-	return val, ok
+	val, exists := (*vc)[key]
+	return val, exists
```
Ajpantuso, line 99:
```suggestion
	return (*vc)[key]
```

pkg/controller/validationscache.go
```diff
@@ -64,35 +72,45 @@ func (vc *validationCache) store(obj client.Object, outcome validations.Validati
...
 }
 
+// objectAlreadyValidated returns if the given object has been validated
```
Ajpantuso, line 102:
```suggestion
// objectAlreadyValidated returns 'true' if the given 'Object'
// has a cached 'ValidationOutcome' with the same 'ResourceVersion'
// (Kubernetes representation of iteration count for a persisted resource).
// If the 'ResourceVersion' of an existing 'Object' is stale the cached
// 'ValidationOutcome' is removed and 'false' is returned. In all other
// cases 'false' is returned.
```

pkg/controller/validationscache.go
```diff
@@ -64,35 +72,45 @@ func (vc *validationCache) store(obj client.Object, outcome validations.Validati
...
+		}
 	}
-	currentResourceVersion := obj.GetResourceVersion()
-	if string(storedResourceVersion) != currentResourceVersion {
-		vc.remove(obj)
-		return false
-	}
-	return true
+
+	return exists
```
Ajpantuso, line 115: This refactor increases nesting depth for the active code path reducing clarity. The 'guard' approach before this change is preferred in this project so this should be reverted.

Ajpantuso (changes requested):
In general we should maintain clear separation between private/public interfaces both in code and documentation. That will reduce duplication of effort if/when this file is re-modularized.

Ajpantuso (approved):
/lgtm

PR #225, opened 2023-01-26
update kubelinter, controller-runtime and also Go versions

tremes:


go.mod
```diff
@@ -1,25 +1,139 @@
 module github.com/app-sre/deployment-validation-operator
 
-go 1.16
+go 1.18
```
tremes, line 3: This will probably not work, because the image seems to have go 1.17

tremes:
Not sure why the prow is still `{"level":"info","msg":"Tagging openshift/boilerplate:image-v2.3.2 into pipeline:root.","time":"2023-01-26T14:29:06Z"}`

tremes:
I am failing to see why the CI is still using the original ` quay.io/app-sre/boilerplate:image-v2.3.2`. Any ideas please @Ajpantuso ?

Ajpantuso:
> I am failing to see why the CI is still using the original ` quay.io/app-sre/boilerplate:image-v2.3.2`. Any ideas please @Ajpantuso ?  

You need to update https://github.com/app-sre/deployment-validation-operator/blob/master/.ci-operator.yaml as well.

tremes:
@Ajpantuso thanks

tremes:
I still don't understand why the `make validate` target is failing here. @Ajpantuso do you know please?

Ajpantuso:
> I still don't understand why the `make validate` target is failing here. @Ajpantuso do you know please? 

I would follow the update instructions for boilerplate https://github.com/openshift/boilerplate/blob/master/README.md first and perform that update in a single PR before this one. As you notes that will only give you go1.17, but it should avoid the dirty repo errors.

tremes:
OK I created https://github.com/app-sre/deployment-validation-operator/pull/227. Thanks

PR #226, opened 2023-01-26
deployment updates to run on 4.12

tremes:
I needed those updates to be able to deploy DVO to my 4.12 OCP cluster. Do we need it? Not sure, but I think we need some reliable way to test & develop easily.

### DIFF
--- a/deploy/openshift/operator.yaml
+++ b/deploy/openshift/operator.yaml
@@ -22,17 +22,15 @@ spec:
         name: deployment-validation-operator
         app: deployment-validation-operator
     spec:
+      securityContext:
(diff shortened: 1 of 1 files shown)


tremes:
This is problematic only when following the https://github.com/app-sre/deployment-validation-operator#manual-installation-without-olm so perhaps we should remove it from the README file and keep only https://github.com/app-sre/deployment-validation-operator#installation-via-olm

Ajpantuso:
`nodeAffinity` for `infra` nodes must not be removed.

tremes:
I created DVO-84 to address this problem This can be closed for now.

PR #246, opened 2023-04-28
DVO-105: extends the logic of matching "app" labels

tremes:
DVO tries to group objects with the same `app` label value. Before this update, the operator tried to read the following paths to get the `app` label value:

- `metadata.labels.app` 
- `spec.selector.matchLabels.app` 

The problem with this approach is that labels can match using `matchExpressions`. An example is:
```
key: app
operator: NotIn
values:
- foo
- bar
```
which means match everything except of resources with `app=foo` or `app=bar`. This is a more effective approach but more complex. 

The problem is that we don't know the set of `app` label values (in a namespace) in advance. So the logic here is to read/process all the "inclusive" label selectors, which builds the set of known `app` label values and then process the other `app` label selectors such as `Exists` and `NotIn`.

### DIFF
--- a/pkg/controller/generic_reconciler.go
+++ b/pkg/controller/generic_reconciler.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
 
+	"github.com/app-sre/deployment-validation-operator/pkg/utils"
 	"github.com/app-sre/deployment-validation-operator/pkg/validations"
 	"github.com/go-logr/logr"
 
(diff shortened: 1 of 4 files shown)


pkg/controller/generic_reconciler.go
```diff
@@ -193,11 +187,38 @@ func (gr *GenericReconciler) groupAppObjects(ctx context.Context,
...
+				// except those that have a corresponding "app" label value.
+				if as.Operator == metav1.LabelSelectorOpNotIn {
+					if as.Values.Has(k) {
+						continue
+					}
+					relatedObjects[k] = append(relatedObjects[k], obj)
+				}
+			}
+		}
+	}
```
tremes, line 218: This is pretty complex, but the assumption here is that the length of `relatedObjects` will be more often greater than length of `appSelectors`. Anothe possible impl would be:
```
	for obj, appSelectors := range postProcessAppSelector {
		for _, as := range appSelectors {
			if as.Operator == metav1.LabelSelectorOpExists {
				for k, _ := range relatedObjects {
					relatedObjects[k] = append(relatedObjects[k], obj)
				}
			} else if as.Operator == metav1.LabelSelectorOpNotIn {
				for k, _ := range relatedObjects {
					if as.Values.Has(k) {
						continue
					}
					relatedObjects[k] = append(relatedObjects[k], obj)
				}
			}
		}
	}
```
but this is iterating over the `relatedObjects` map multiple times. I can provide a benchmark test if anyone wants to experiment.

tremes:
Closing this one in favor of https://github.com/app-sre/deployment-validation-operator/pull/252

PR #252, opened 2023-05-29
DVO-105: extends the logic of matching "app" labels

tremes:
DVO tries to group objects with the same app label value. Before this update, the operator tried to read the following paths to get the app label value:

- `metadata.labels.app`
- `spec.selector.matchLabels.app`

The problem with this approach is that labels can match using matchExpressions. An example is:
```
key: app
operator: NotIn
values:
- foo
- bar
```
which means match everything except of resources with app=foo or app=bar. This is a more effective approach but more complex.

The problem is that we don't know the set of app label values (in a namespace) in advance. So the logic here is to sort the found `GroupVersionKinds` according their group (this adds the `apps/v1` group between the first in the slice) and then process the `app` label selectors.


ncaak (approved):
/lgtm

PR #253, opened 2023-06-02
DVO-110: Move disable checks to configmap

ncaak:
### summary
* Extract the hardcoded disabled checks to a configmap default configuration (yaml)
* Fix ConfigMap watcher which in the future will take care of dynamic configuration


pkg/controller/configmap_watcher.go
```diff
@@ -14,13 +16,28 @@ import (
...
+
 type ConfigMapWatcher struct {
 	clientset kubernetes.Interface
-	ch        chan struct{} // TODO - TBD configmap struct
+	checks    KubeLinterChecks
+	ch        chan config.Config
 }
 
-var configMapName = "deployment-validation-operator-config"
+var configMapName = "deployment-validation-operator-custom-config"
```
ncaak, line 38: Maintaining compatibility on previous ConfigMap name. It's already mounted in a volume and it's redundant to get it from the API.

tremes:
@ncaak it looks like it would be good to update the `README` file too.

pkg/controller/configmap_watcher.go
```diff
@@ -36,16 +53,17 @@ func NewConfigMapWatcher(cfg *rest.Config) (ConfigMapWatcher, error) {
...
-func (cmw *ConfigMapWatcher) GetStaticDisabledChecks(ctx context.Context) ([]string, error) {
+// GetStaticKubelinterConfig returns the ConfigMap's checks configuration
+func (cmw *ConfigMapWatcher) GetStaticKubelinterConfig(ctx context.Context) (config.Config, error) {
+	var cfg config.Config
+
 	cm, err := cmw.clientset.CoreV1().
 		ConfigMaps(configMapNamespace).Get(ctx, configMapName, v1.GetOptions{})
 	if err != nil {
-		return []string{}, fmt.Errorf("gathering starting configmap: %w", err)
+		return cfg, fmt.Errorf("gathering starting configmap: %w", err)
```
tremes, line 63: nit. This error messages seems to me a bit strange to be honest. What about something like getting initial configuration or something like that?

main.go
```diff
@@ -235,3 +244,19 @@ func kubeClientQPS() (float32, error) {
...
+
+// getCustomConfig returns a custom Kube-Linter configuration
+// it retrieves data from a ConfigMap if there is any valid
+func getCustomConfig(cfg *rest.Config) (*kubelinterCfg.Config, error) {
+	cmw, err := controller.NewConfigMapWatcher(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("getting kube-linter checks configuration from configmap: %w", err)
+	}
+
+	klCfg, err := cmw.GetStaticKubelinterConfig(context.Background())
```
tremes, line 256: I think we should ideally pass a context instance to the `getCustomConfig` and use it here. WDYT?

pkg/controller/configmap_watcher.go
tremes: I think I don't understand the reason behind this rename.

pkg/controller/configmap_watcher.go
tremes: Also should we return empty `Config` struct in this case?

pkg/controller/configmap_watcher.go
ncaak: Currently default ConfigMap is part of the Deployment and we cannot change it. So, this adds a custom ConfigMap which data will be prioritized over the default one (immutable). And I added `custom` to the ConfigMap name to not overlap the basic, default one.

pkg/controller/configmap_watcher.go
```diff
@@ -57,14 +75,17 @@ func (cmw *ConfigMapWatcher) StartInformer(ctx context.Context) error {
...
 			newCm := newObj.(*apicorev1.ConfigMap)
 
-			fmt.Printf("oldCm: %v\n", oldCm)
 			fmt.Printf("ConfigMap updated: %s/%s\n", newCm.Namespace, newCm.Name)
 
-			// TODO - Validate new configmap
-			cmw.ch <- struct{}{}
+			cfg, err := cmw.getKubeLinterConfig(newCm.Data[configMapDataAccess])
+			if err != nil {
+				fmt.Printf("Error: unmarshalling configmap data: %s", err.Error())
```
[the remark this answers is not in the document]
ncaak: Ok, I'll take a look and reuse logger for error messages as well.

pkg/controller/configmap_watcher.go
ncaak: > nit. This error messages seems to me a bit strange to be honest. What about something like getting initial configuration or something like that?

Taking advantage on before comment, I'll revamp new log messages as well.

> Also should we return empty Config struct in this case?

That config returned is not being used anywhere. I usually send the most information I have to possible debugging in the future if something goes wrong. If you think is a better idea to send an empty object, I don't have a strong opinion on this.

main.go
ncaak: I thought about it, but there was no context used on the upstream. I can send start it there and then send it as argument :+1:

tremes: There's `if err := mgr.Start(signals.SetupSignalHandler()); err != nil {` in the main function, when `signals.SetupSignalHandler())` returns a context, so I think you can take it from there.

pkg/validations/validation_engine.go
```diff
@@ -52,13 +52,17 @@ func fileExists(filename string) bool {
 }
 
 func (ve *validationEngine) LoadConfig(path string) error {
-	v := viper.New()
-
 	if !fileExists(path) {
 		log.Info(fmt.Sprintf("config file %s does not exist. Use default configuration", path))
-		path = ""
+		// legacy disabled checks
+		ve.config.Checks.Exclude = getDisabledChecks()
```
tremes, line 58: I think we can enable everything when the configmap doesn't exist. We will rely on its existence in the OSD clusters.

README.md
```diff
@@ -99,29 +99,27 @@ The metrics generated by DVO can be scraped by anything that understands prometh
...
 
-```
-Location of file 
-../pkg/validations/validation_engine.go 
-```
+To disable all checks, set the `doNotAutoAddDefaults` property to `true`. If you only want to disable individual checks, include them as a collection in the `exclude` property and leave `doNotAutoAddDefaults` with a value of `false`
+
+The `exclude` property takes precedence over the `include` property. If a particular check is in both collections, it will be excluded by default.
+
+The `exclude` property can work in conjunction with `addAllBuiltIn` set to `true` in a blacklisting fashion. All checks will be triggered and only the checks passed in `exclude` will be ignored.
```
tremes, line 122: Maybe we could also say what the default configuration is (when the configmap doesn't exist).

pkg/validations/validation_engine.go
ncaak: I was only maintaining the compatibility. Previous version was always excluding those checks, but if you think it'd be a good idea to enable everything (but the incompatible ones), I'll change this and the default configmaps yamls on the repo.

tremes: As discussed yesterday, we don't want to enable everything. Marking as resolved.

tremes:
I reviewed and experimented a little bit with this one. Thanks.
/lgtm

PR #254, opened 2023-06-12
DVO-101: add a new template for HyperShift SelectorSyncSet

tremes:


hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,214 @@
...
+      annotations:
+        component-display-name: ${DISPLAY_NAME}
+        component-name: ${REPO_NAME}
+      labels:
+        managed.openshift.io/gitHash: ${IMAGE_TAG}
+        managed.openshift.io/gitRepoName: ${REPO_NAME}
+        managed.openshift.io/osd: "true"
+      name: deployment-validation-operator
+    spec:
+      clusterDeploymentSelector:
```
Nikokolas3270, line 40: MC cluster selector should be:
```yaml
      clusterDeploymentSelector:
        matchExpressions:
          - key: ext-hypershift.openshift.io/cluster-type
            operator: In
            values:
              - management-cluster
          - key: api.openshift.com/fedramp
            operator: NotIn
            values:
              - "true"
      resourceApplyMode: Sync
```

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,214 @@
...
+            deployment-validation-operator-config.yaml: |-
+              checks:
+                # if doNotAutoAddDefaults is true, default checks are not automatically added.
+                doNotAutoAddDefaults: false
+
+                # addAllBuiltIn, if set, adds all built-in checks. This allows users to
+                # explicitly opt-out of checks that are not relevant using Exclude.
+                # Takes precedence over doNotAutoAddDefaults, if both are set.
+                addAllBuiltIn: true
+        - apiVersion: policy.open-cluster-management.io/v1
```
Nikokolas3270, line 153: All above listed objects should be wrapped in this object.
See:
https://github.com/openshift/hypershift-dataplane-metrics-forwarder/blob/d400ebc5166c0b3f0167973558b11fd7a39487b9/hack/templates/00-hypershift-dataplane-metrics-forwarder.SelectorSyncSet.yaml.tmpl#L123

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,214 @@
...
+          data:
+            deployment-validation-operator-config.yaml: |-
+              checks:
+                # if doNotAutoAddDefaults is true, default checks are not automatically added.
+                doNotAutoAddDefaults: false
+
+                # addAllBuiltIn, if set, adds all built-in checks. This allows users to
+                # explicitly opt-out of checks that are not relevant using Exclude.
+                # Takes precedence over doNotAutoAddDefaults, if both are set.
+                addAllBuiltIn: true
```
tremes, line 152: Update the configmap

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,214 @@
...
+    kind: SelectorSyncSet
+    metadata:
+      annotations:
+        component-display-name: ${DISPLAY_NAME}
+        component-name: ${REPO_NAME}
+      labels:
+        managed.openshift.io/gitHash: ${IMAGE_TAG}
+        managed.openshift.io/gitRepoName: ${REPO_NAME}
+        managed.openshift.io/osd: "true"
+      name: deployment-validation-operator
```
Nikokolas3270, line 38: Could you rename it `deployment-validation-operator-hypershift` to avoid a conflict with the existing `deployment-validation-operator` SelectorSyncSet

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,214 @@
...
+                        hypershift.openshift.io/hosted-control-plane: "true"
+                    object-templates:
+                      - complianceType: MustHave
+                        objectDefinition:
+                          apiVersion: package-operator.run/v1alpha1
+                          kind: Package
+                          metadata:
+                            name: deployment-validation-operator
+                          spec:
+                            image: ${REGISTRY_IMG}@${IMAGE_DIGEST}
```
Nikokolas3270, line 186: Replace this objects with all the objects you listed before the policy (Namespace, Subscription, Service...)

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,214 @@
...
+                  kind: ConfigurationPolicy
+                  metadata:
+                    name: deployment-validation-operator
+                  spec:
+                    evaluationInterval:
+                      compliant: 2h
+                      noncompliant: 45s
+                    namespaceSelector:
+                      matchLabels:
+                        hypershift.openshift.io/hosted-control-plane: "true"
```
Nikokolas3270, line 177: That's not needed, we do not want to deploy the objects in `object-templates` in the HCP namespace on the MC cluster.

Nikokolas3270:
Please note that the following `SelectorSyncSet` template will need to be deployed through `app-interface` repository.
Indeed, as discussed, `managed-cluster-config` repository is only for deploying objects/resources on OSD/ROSA clusters used by the customers. Hypershift management clusters (MCs) are OSD/ROSA clusters but they are not customer clusters.
Take a look at this file to know how to deploy a template on hive clusters with that scope in mind:
https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/osd-operators/cicd/saas/saas-hypershift-platform-rhobs-rules.yaml#L45
Of course, we can do the `app-interface` MR together :)

Nikokolas3270:
Hello Tomas, the PR now looks fine to me.
Lets perform some tests on a guinea pig management cluster (MC) tomorrow for the `Policy` and the related objects to see if the objects contained in the policy correctly get deployed on the hosted clusters.

hack/olm-registry/hypershift-template.yaml
Nikokolas3270: Actually the selector should be:
```
      clusterDeploymentSelector:
        matchLabels:
          ext-hypershift.openshift.io/cluster-type: service-cluster
```
We need to deploy the policy, the placement rule and the binding rule on the service cluster. Those objects are then duplicated by the service cluster on each klusterlet namespace for each hosted cluster on each management cluster.

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,219 @@
...
+          metadata:
+            name: placement-deployment-validation-operator
+            namespace: openshift-acm-policies
+          spec:
+            clusterConditions:
+            - status: "True"
+              type: ManagedClusterConditionAvailable
+            clusterSelector:
+              matchExpressions:
+                - key: hypershift.open-cluster-management.io/management-cluster
```
tremes, line 203: This key is also wrong. It should be `hypershift.open-cluster-management.io/hosted-cluster`

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,219 @@
...
+                  kind: ConfigurationPolicy
+                  metadata:
+                    name: deployment-validation-operator
+                  spec:
+                    evaluationInterval:
+                      compliant: 2h
+                      noncompliant: 45s
+                    object-templates:
+                      - complianceType: MustHave
+                        objectDefinition:
```
tremes, line 76: We need to repeat this for every resource below, because this `objectDefinition` doesn't have array-like value.

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,219 @@
...
+                            namespace: openshift-deployment-validation-operator
+                          spec:
+                            podSelector: {}
+                            ingress:
+                              - from:
+                                - namespaceSelector:
+                                    matchLabels:
+                                      name: openshift-operator-lifecycle-manager
+                          policyTypes:
+                            - Ingress
```
tremes, line 151: Wrong indentation here.

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,219 @@
...
+                            namespace: openshift-deployment-validation-operator
+                          spec:
+                            podSelector: {}
+                            ingress:
+                              - from:
+                                - namespaceSelector:
+                                    matchLabels:
+                                      name: openshift-insights
+                          policyTypes:
+                            - Ingress
```
tremes, line 137: Wrong indentation here.

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,219 @@
...
+                      noncompliant: 45s
+                    object-templates:
+                      - complianceType: MustHave
+                        objectDefinition:
+                        - apiVersion: package-operator.run/v1alpha1
+                          kind: Package
+                          metadata:
+                            name: deployment-validation-operator
+                          spec:
+                            image: ${REGISTRY_IMG}@${IMAGE_DIGEST}
```
tremes, line 82: I guess the Package is not needed.

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,220 @@
...
+                  metadata:
+                    name: deployment-validation-operator
+                  spec:
+                    evaluationInterval:
+                      compliant: 2h
+                      noncompliant: 45s
+                    object-templates:
+                      - complianceType: MustHave
+                        objectDefinition:
+                        - apiVersion: v1
```
Nikokolas3270, line 70: No dash here :)

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,220 @@
...
+                      - complianceType: MustHave
+                        objectDefinition:
+                          apiVersion: operators.coreos.com/v1alpha1
+                          kind: Subscription
+                          metadata:
+                            name: deployment-validation-operator
+                            namespace: openshift-deployment-validation-operator
+                          spec:
+                            channel: ${CHANNEL}
+                            config:
```
Nikokolas3270, line 110: Affinity of the operator to infra nodes has to be removed.
Here the code to add under `config`:
```
config:
  affinity:
    nodeAffinity: {}
```

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,220 @@
...
+                              env:
+                              - name: "NAMESPACE_IGNORE_PATTERN"
+                                value: "^(openshift.*|kube-.*|default|dedicated-admin|redhat-.*|acm|addon-dba-operator|codeready-.*|prow)$"
+                            name: deployment-validation-operator
+                            source: deployment-validation-operator-catalog
+                            sourceNamespace: openshift-deployment-validation-operator
```
Nikokolas3270, line 116: To be changed to the `openshift-marketplace` namespace.

Nikokolas3270 (commented):
Hello,
Some mofications are still required to successfully install the DVO operator.
Cheers,
     Nicolas

hack/olm-registry/hypershift-template.yaml
tremes: I am not sure I understand here. So the affinity default value is not an empty object/type `{}`?

tremes: ohh so this is to override the https://github.com/app-sre/deployment-validation-operator/blob/master/deploy/openshift/operator.yaml#L30 right? Then it makes sense.....probably :)

Nikokolas3270: Yes this overrides that config with an empty (`{}`) config.

hack/olm-registry/hypershift-template.yaml
Nikokolas3270: @tremes: you can actually keep installing the catalog in the `openshift-deployment-validation-operator` namespace. So please ignore above comment.

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,220 @@
...
+                            publisher: Red Hat
+                      - complianceType: MustHave
+                        objectDefinition:
+                          apiVersion: operators.coreos.com/v1
+                          kind: OperatorGroup
+                          metadata:
+                            name: deployment-validation-operator-og
+                            namespace: openshift-deployment-validation-operator
+                            annotations:
+                              olm.operatorframework.io/exclude-global-namespace-resolution: "true"
```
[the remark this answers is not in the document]
Nikokolas3270: @tremes: removing this annotation was only making sense when deploying the catalog in the `openshift-marketplace` namespace. It is better to keep this annotation as the catalog and the operator pods are deployed in the same namespace... which is the case when deploying the catalog in the `openshift-deployment-validation-operator` namespace. So please ignore above comment.

hack/olm-registry/hypershift-template.yaml
```diff
@@ -0,0 +1,222 @@
...
+                            namespace: openshift-deployment-validation-operator
+                          spec:
+                            podSelector: {}
+                            ingress:
+                              - from:
+                                - namespaceSelector:
+                                    matchLabels:
+                                      name: openshift-operator-lifecycle-manager
+                            policyTypes:
+                              - Ingress
```
Nikokolas3270, line 150: Spec to be changed to:
```
                          spec:
                            podSelector:
                              matchLabels:
                                olm.catalogSource: deployment-validation-operator-catalog
                            ingress:
                              - {}
                            policyTypes:
                              - Ingress
```
OLM pod is indeed running in the HCP namespace (so not in the `openshift-operator-lifecycle-manager` ns) on the MC cluster and is not visible in the dataplane.

Solution here is:
- To allow all external traffic.
- But only to the catalog pod.

tremes: Update thank you

ncaak (approved):
/lgtm

PR #256, opened 2023-06-23
Improve CSV generation script for OCP clusters

JoaoFula:
The no-nodeAffinity option allows us to use the script to generate CSV files that can be used for OCP installations.
The other change regards the output-dir variable where previously, you'd have to have the output-dir previously created otherwise the script would fail.

### DIFF
--- a/hack/generate-operator-bundle-contents.py
+++ b/hack/generate-operator-bundle-contents.py
@@ -24,6 +24,9 @@ optional.add_argument( '-h', '--help', action='help', default=argparse.SUPPRESS,
 optional.add_argument('-r', '--replaces', help='Replaces version', type=str)
 optional.add_argument('-s','--skip', action='append',
         help='Skips version (can be specified multiple times)')
+optional.add_argument('--no-nodeAffinity', '--no-nodeAffinity',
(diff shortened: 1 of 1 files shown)


hack/generate-operator-bundle-contents.py
```diff
@@ -82,5 +87,6 @@
 
 csv_filename = pathlib.Path(args.output_dir) / \
     f'{args.name}.v{args.current_version}.clusterserviceversion.yaml'
+os.makedirs(os.path.dirname(csv_filename), exist_ok=True)
```
tremes, line 90: what is this good for please? It's not obvious to me on Friday.... :)

JoaoFula: Basically when we pass the args.output_dir, if the directory isn't already created (for example, bundle/manifests), the script fails because it cannot find the path. This makes sure that if the path doesn't exist, it creates it before doing the same with the file.

tremes:
I verified it locally and it works as expected. I think it's acceptable solution for now to be able to generate testing OLM bundle for on-prem OCP clusters. 
/lgtm

PR #258, opened 2023-06-29
Prefer scheduling registry pod onto infra nodes

anispate:
This commit specifies .spec.grpcPodConfig.nodeSelector and .spec.grpdPodConfig.tolerations to schedule this operator's registry pod to infra nodes when possible.

[OSD-6629](https://issues.redhat.com//browse/OSD-6629)


tremes:
Hi @anispate Thanks for the PR. What happens please when there's no infra node in the cluster? Will it deploy OK or will it fail?

anispate:
If there is no infra node in the cluster then it won't get deployed.

ncaak (approved):

PR #259, opened 2023-06-29
DVO-128: enable non-isolated Pod check & enhance group matching again

tremes:
This enables the existing `nonisolated-pod` kube-linter check. I realized (when working & playing with this one) that the labels & selectors logic we have for grouping the objects is very poor. I created two new tasks which should be addressed:

- https://issues.redhat.com/browse/DVO-132
- https://issues.redhat.com/browse/DVO-133


tremes:
closing in favor of https://github.com/app-sre/deployment-validation-operator/pull/262

PR #260, opened 2023-07-03
DVO-132: empty selector matches all the resource in the corresponding

tremes:
namespace

If the resource has empty "selector" defined then it matches all the objects in the corresponding namespace.


ncaak (approved):
/lgtm

PR #262, opened 2023-07-04
DVO-128: enable non-isolated Pod check & DVO-133 fix and enhance grou…

tremes:
…p matching

This should address two tasks/issues:
- [DVO-128](https://issues.redhat.com//browse/DVO-128) - enable non-isolated Pod kubelinter check
- [DVO-133](https://issues.redhat.com//browse/DVO-133) - selectors are not ANDed - I tried to reimplement the logic of reading labels as well as selectors in more robust way. Now it tries to use types and functions provided by the Kubernetes `apimachinery` (especially the `LabelSelector` type wrapping all the requirements) as much as possible. Test have been extended as well.


pkg/utils/selector.go
```diff
@@ -0,0 +1,118 @@
...
+	return nil
+}
+
+// readMatchExpressions tries to parse known paths for the "matchExpressions" attribute in the
+// unstructured object input
+func readMatchExpressions(object *unstructured.Unstructured) []metav1.LabelSelectorRequirement {
+	matchExpressionsRequirements := []metav1.LabelSelectorRequirement{}
+	for _, path := range []manifestPath{selectorMatchExpressions, podSelectorMatchExpressions} {
+		pathExists := pathExistsAsSlice(object, path)
+		if pathExists {
```
ncaak, line 59: nit: this function could be less complex just inverting the conditional.
Instead of:
```
if true {
    // stuff with depth
}
```
You can just write:
```
if !pathExists {
    continue
}
// stuff here
```

It's not big deal but with a function like this (that contains 4 depth tabs) it could make the code easier to read for a cheap fast exit.

tremes: Yeah sure. OK. Thanks

PR #263, opened 2023-08-09
DVO-111: Dynamic validation engine

ncaak:
### summary

* Allow the project to change the configuration in execution time, updating the validations with the new configuration updated on the ConfigMap

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,7 +17,7 @@ import (
 	"github.com/app-sre/deployment-validation-operator/internal/options"
 	"github.com/app-sre/deployment-validation-operator/pkg/configmap"
 	"github.com/app-sre/deployment-validation-operator/pkg/controller"
-	dvo_prom "github.com/app-sre/deployment-validation-operator/pkg/prometheus"
+	dvoProm "github.com/app-sre/deployment-validation-operator/pkg/prometheus"
 	"github.com/app-sre/deployment-validation-operator/pkg/validations"
 	"github.com/app-sre/deployment-validation-operator/version"
 
@@ -97,14 +97,14 @@ func setupManager(log logr.Logger, opts options.Options) (manager.Manager, error
 
 	log.Info("-> Initialize Prometheus Registry")
 
-	reg, err := dvo_prom.SetupRegistry()
+	reg, err := dvoProm.SetupRegistry()
 	if err != nil {
 		return nil, fmt.Errorf("initializing Prometheus server: %w", err)
 	}
 
 	log.Info(fmt.Sprintf("-> Initialize Prometheus metrics endpoint on %q", opts.MetricsEndpoint()))
 
-	srv, err := dvo_prom.NewServer(reg, opts.MetricsPath, fmt.Sprintf(":%d", opts.MetricsPort))
+	srv, err := dvoProm.NewServer(reg, opts.MetricsPath, fmt.Sprintf(":%d", opts.MetricsPort))
 	if err != nil {
 		return nil, fmt.Errorf("initializing metrics server: %w", err)
 	}
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 	dvoProm "github.com/app-sre/deployment-validation-operator/pkg/prometheus"
 	"github.com/app-sre/deployment-validation-operator/pkg/validations"
 	"github.com/app-sre/deployment-validation-operator/version"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/go-logr/logr"
 	osappsv1 "github.com/openshift/api/apps/v1"
@@ -97,9 +98,10 @@ func setupManager(log logr.Logger, opts options.Options) (manager.Manager, error
 
 	log.Info("-> Initialize Prometheus Registry")
 
-	reg, err := dvoProm.SetupRegistry()
+	reg := prometheus.NewRegistry()
+	metrics, err := dvoProm.PreloadMetrics(reg)
 	if err != nil {
-		return nil, fmt.Errorf("initializing Prometheus server: %w", err)
+		return nil, fmt.Errorf("preloading kube-linter metrics: %w", err)
 	}
 
 	log.Info(fmt.Sprintf("-> Initialize Prometheus metrics endpoint on %q", opts.MetricsEndpoint()))
@@ -126,7 +128,7 @@ func setupManager(log logr.Logger, opts options.Options) (manager.Manager, error
 
 	log.Info("-> Initialize Validation Engine")
 
-	validationEngine, err := validations.NewEngine(opts.ConfigFile, cmWatcher)
+	validationEngine, err := validations.NewEngine(opts.ConfigFile, cmWatcher, metrics)
 	if err != nil {
 		return nil, fmt.Errorf("initializing validation engine: %w", err)
 	}
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ import (
 	apis "github.com/app-sre/deployment-validation-operator/api"
 	dvconfig "github.com/app-sre/deployment-validation-operator/config"
 	"github.com/app-sre/deployment-validation-operator/internal/options"
+	"github.com/app-sre/deployment-validation-operator/pkg/configmap"
 	"github.com/app-sre/deployment-validation-operator/pkg/controller"
 	dvo_prom "github.com/app-sre/deployment-validation-operator/pkg/prometheus"
 	"github.com/app-sre/deployment-validation-operator/pkg/validations"
@@ -85,48 +86,23 @@ func setupManager(log logr.Logger, opts options.Options) (manager.Manager, error
 		return nil, fmt.Errorf("getting config: %w", err)
 	}
 
-	log.Info("Initialize Scheme")
-
-	scheme, err := initializeScheme()
-	if err != nil {
-		return nil, fmt.Errorf("initializing scheme: %w", err)
-	}
-
 	log.Info("Initialize Manager")
 
-	mgrOpts, err := getManagerOptions(scheme, opts)
-	if err != nil {
-		return nil, fmt.Errorf("getting manager options: %w", err)
-	}
-
-	mgr, err := manager.New(cfg, mgrOpts)
+	mgr, err := initManager(log, opts, cfg)
 	if err != nil {
 		return nil, fmt.Errorf("initializing manager: %w", err)
 	}
 
-	if err := mgr.AddHealthzCheck("health", healthz.Ping); err != nil {
-		return nil, fmt.Errorf("adding healthz check: %w", err)
-	}
-
-	if err := mgr.AddReadyzCheck("check", healthz.Ping); err != nil {
-		return nil, fmt.Errorf("adding readyz check: %w", err)
-	}
-
 	log.Info("Registering Components")
 
-	discoveryClient, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
-	if err != nil {
-		return nil, fmt.Errorf("initializing discovery client: %w", err)
-	}
-
-	log.Info("Initializing Prometheus Registry")
+	log.Info("-> Initialize Prometheus Registry")
 
 	reg, err := dvo_prom.SetupRegistry()
 	if err != nil {
 		return nil, fmt.Errorf("initializing Prometheus server: %w", err)
 	}
 
-	log.Info(fmt.Sprintf("Initializing Prometheus metrics endpoint on %q", opts.MetricsEndpoint()))
+	log.Info(fmt.Sprintf("-> Initialize Prometheus metrics endpoint on %q", opts.MetricsEndpoint()))
 
 	srv, err := dvo_prom.NewServer(reg, opts.MetricsPath, fmt.Sprintf(":%d", opts.MetricsPort))
 	if err != nil {
@@ -137,13 +113,9 @@ func setupManager(log logr.Logger, opts options.Options) (manager.Manager, error
 		return nil, fmt.Errorf("adding metrics server to manager: %w", err)
 	}
 
-	log.Info("Initializing Validation Engine")
-	validationEngine, err := validations.NewEngine(opts.ConfigFile)
-	if err != nil {
-		return nil, fmt.Errorf("initializing validation engine: %w", err)
-	}
+	log.Info("-> Initialize ConfigMap watcher")
 
-	cmWatcher, err := controller.NewConfigMapWatcher(cfg)
+	cmWatcher, err := configmap.NewConfigMapWatcher(cfg)
 	if err != nil {
 		return nil, fmt.Errorf("initializing configmap watcher: %w", err)
 	}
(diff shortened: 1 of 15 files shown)


pkg/prometheus/prometheus.go
```diff
@@ -80,6 +82,52 @@ func getRouter(registry Registry, path string) (*http.ServeMux, error) {
 	return mux, nil
 }
 
+// GetRegistry returns a fully configured Prometheus registry with metrics based on Kube Linter validations
+func GetRegistry() (*prometheus.Registry, error) {
```
tremes, line 86: nit: Is `GetRegistry` a good name here? Perhaps it looks more like initialization or setup of registry...

pkg/prometheus/prometheus.go
```diff
@@ -80,6 +82,52 @@ func getRouter(registry Registry, path string) (*http.ServeMux, error) {
...
+
+	checks, err := validations.GetAllNamesFromRegistry(reg)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, checkName := range checks {
+		metric, err := setupMetric(reg, checkName)
+		if err != nil {
+			return nil, fmt.Errorf("unable to create metric for check %s", checkName)
```
tremes, line 102: Hmm so I guess this means that if the `checkName` is not known (or maybe just mispelled. I am looking at the line 115 below) then the DVO init will fail with error right? Do we want such behavior?

pkg/controller/configmap_watcher.go
```diff
@@ -54,6 +54,7 @@ func NewConfigMapWatcher(cfg *rest.Config) (ConfigMapWatcher, error) {
 	return ConfigMapWatcher{
 		clientset: clientset,
 		logger:    log.Log.WithName("ConfigMapWatcher"),
+		ch:        make(chan config.Config),
```
tremes, line 57: Another approach could be to just have it for an empty struct{} to notify about the event and then obtain the config on the listener side.

pkg/controller/generic_reconciler.go
```diff
@@ -94,6 +105,8 @@ func (gr *GenericReconciler) AddToManager(mgr manager.Manager) error {
 
 // Start validating the given object kind every interval.
 func (gr *GenericReconciler) Start(ctx context.Context) error {
+	go gr.ConfigChanged(ctx)
```
tremes, line 108: Just an idea to consider here. What about moving the listener part directly to the validation engine and then update registry directly there?

pkg/prometheus/prometheus.go
ncaak: Yes, this will change on next refactor. I "copied" before method but will definitely change it to a more suitable name.

pkg/prometheus/prometheus.go
ncaak: This was an existing logic within the validationEngine package where an error on spelling a validation name will throw an error.

I agree this could be handled better with a logline and skipping the error. I'll do some testing and come back with a change as I'm refactoring the function where this behaviour was originally.

tremes: Yes you're right, but previously the values were hardcoded,  but now they can be changed as the user wishes, which means it is more error-prone in my opinion.

pkg/controller/generic_reconciler.go
tremes: Well you would need to add the `Start` method to the validation engine and then register it in the manager as well.

pkg/prometheus/prometheus.go
ncaak: Yes, good point. I'll do some testing and refactor this with a different approach.
Since we can't block ConfigMap update, I'll validate the structure (currently just marshalling into a struct) and then allow errors in the own configuration (logline informing about the wrong spelling/configuration and skipping any errors that could break the system.

pkg/configmap/configmap_watcher.go
```diff
@@ -1,8 +1,9 @@
-package controller
+package configmap
```
tremes, line 1: Wouldn't be better to call the package just `config` ?

pkg/configmap/configmap_watcher.go
```diff
@@ -79,6 +81,11 @@ func (cmw *ConfigMapWatcher) StartInformer(ctx context.Context) error {
 		UpdateFunc: func(oldObj, newObj interface{}) {
 			newCm := newObj.(*apicorev1.ConfigMap)
 
+			// This is sometimes triggered even if no change was due to the ConfigMap
+			if configMapName != newCm.ObjectMeta.Name || reflect.DeepEqual(oldObj, newObj) {
```
tremes, line 85: Yes I think the informer is resynced periodically....is it an issue?

main.go
```diff
@@ -86,68 +86,69 @@ func setupManager(log logr.Logger, opts options.Options) (manager.Manager, error
...
-	mgr, err := manager.New(cfg, mgrOpts)
+	reg, err := dvoProm.SetupRegistry()
 	if err != nil {
-		return nil, fmt.Errorf("initializing manager: %w", err)
+		return nil, fmt.Errorf("initializing Prometheus server: %w", err)
 	}
 
-	if err := mgr.AddHealthzCheck("health", healthz.Ping); err != nil {
-		return nil, fmt.Errorf("adding healthz check: %w", err)
+	log.Info(fmt.Sprintf("-> Initialize Prometheus metrics endpoint on %q", opts.MetricsEndpoint()))
```
tremes, line 105: I think `->` is not necessary. I think the better way to distinguish it would be to use a log with different name or perhaps different log level. I don't know `zap` much....

pkg/prometheus/prometheus.go
```diff
@@ -80,6 +82,56 @@ func getRouter(registry Registry, path string) (*http.ServeMux, error) {
...
+		}
+
+		if err := prom.Register(metric); err != nil {
```
tremes, line 105: I think we register too many metrics too early here, because if I am not mistaken the config is loaded afterwards (when initializing the `validationEngine`, which creates the metrics again, but never registers them).

main.go
ncaak: I was using that symbol as a bullet for the log, to have a minimal appraise of what was going on.
e.g.
`2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	Load KubeConfig
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	Initialize Manager
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	-> Initialize Scheme
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	-> Getting Manager Options
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	-> Adding Healthz and Readyz checks
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	Registering Components
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	-> Initialize Prometheus Registry
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	-> Initialize Prometheus metrics endpoint on "http://0.0.0.0:8383/metrics"
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	-> Initialize ConfigMap watcher
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	-> Initialize Validation Engine
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	-> Initialize Reconciler
2023-08-22T12:31:11+02:00	INFO	DeploymentValidation	Starting Manager`

pkg/configmap/configmap_watcher.go
ncaak: I had a lint complaint (fixed already) to rename the structure inside the file as Watcher, instead of ConfigmapWatcher.

I think `configmap.Watcher` is way more specific and readable than`config.Watcher`.

Also, there is some other packages imported inside the project with the generic name `config`.

But I don't have strong opinion on this, if you think it's better to call it config, it's a minimal effortless change.

pkg/configmap/configmap_watcher.go
ncaak: Apparently this is expected behaviour, so I've added the reflect library to compare. After the first time it comes in with both old and new object being the same, it causes a return and this avoids further Informer updates with the same change.

But I wanted to "fix" a constant recurrent fake update (because there is no new information) to be updating Validation Engine.

pkg/prometheus/prometheus.go
ncaak: After reviewing this and testing it, this will only trigger an error if the validation checkname is wrong at software start.

If this comes from an update on the ConfigMap, the error will be ignored and logged by the code.

this line :point_down: 
https://github.com/app-sre/deployment-validation-operator/pull/263/files#diff-48a7386f3f126ad7b562cfde8d8b4ac24bf7e080225fc7afeee7b35eb2ed3161R198

pkg/controller/generic_reconciler.go
ncaak: This change was implemented.

pkg/controller/configmap_watcher.go
ncaak: I missed this comment.

Now this structure exists as a field of the Validation Engine structure. This approach would work by exporting the configuration within the structure and making the Validation Engine logic to retrieve it.

Is there any security concern or convention for this to change to an empty struct? (btw I think this was the previous approach when the watcher was coded and merged)

pkg/configmap/configmap_watcher.go
tremes: OK. Yes `configmap.Watcher` is probably better. It was just a suggestion...you can keep it as it is.

main.go
tremes: Yes I know but I think this is lost when you have the log formatted in JSON.

ncaak: I was checking Zap library but nothing like this except for changing the name, and then you would have to change it back. It looks like a lot of work only for a little change.

Rolling back. :+1:

main.go
```diff
@@ -86,68 +87,70 @@ func setupManager(log logr.Logger, opts options.Options) (manager.Manager, error
...
-		return nil, fmt.Errorf("adding readyz check: %w", err)
+	if err := mgr.Add(srv); err != nil {
+		return nil, fmt.Errorf("adding metrics server to manager: %w", err)
 	}
 
-	log.Info("Registering Components")
+	log.Info("Initialize ConfigMap watcher")
 
-	discoveryClient, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
+	cmWatcher, err := configmap.NewConfigMapWatcher(cfg)
```
tremes, line 120: It would be nice to rename the constructor as well - `NewWatcher`

pkg/validations/validation_engine.go
```diff
@@ -2,43 +2,79 @@ package validations
...
-	_ "golang.stackrox.io/kube-linter/pkg/templates/all"
+	_ "golang.stackrox.io/kube-linter/pkg/templates/all" // nolint:golint
 
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/spf13/viper"
 )
 
-var engine validationEngine
+var engine ValidationEngine
```
tremes, line 27: Let's not do it in this PR, but it would be good to remove it in the future.

pkg/validations/utils.go
```diff
@@ -4,24 +4,75 @@ import (
...
 		prometheus.GaugeOpts{
-			Name:        fmt.Sprintf("%s_%s", strings.ReplaceAll(config.OperatorName, "-", "_"), name),
-			Help:        help,
-			ConstLabels: constLabels,
-		},
-		labelNames,
-	)
+			Name: strings.ReplaceAll(
+				fmt.Sprintf("%s_%s", config.OperatorName, check.Name),
+				"-", "_"),
```
tremes, line 69: nit: it would be nice to move this to a new variable

pkg/validations/validation_engine.go
ncaak: You mean the "global variable" approach or the export of the structure?

tremes: yes the global variable "hack" :)

pkg/controller/configmap_watcher.go
tremes: Not sure about security concern. My experience is with channels with `struct{}` or perhaps string - i.e keep them lightweight. Let's keep this as it is and focus on more important problems here. We can always update in the future.

pkg/configmap/configmap_watcher.go
```diff
@@ -0,0 +1,177 @@
...
+		List(context.Background(), v1.ListOptions{})
+	if err != nil {
+		return "", fmt.Errorf("getting deployments from clientset: %w", err)
+	}
+
+	for _, deployment := range deployments.Items {
+		if deployment.GetName() == dvoConfig.OperatorName {
+			return deployment.GetNamespace(), nil
+		}
+	}
```
tremes, line 173: I would have other suggestion (sorry maybe I am too picky). There's `POD_NAME` env variable available for the DVO container. We can do `os.Getenv("POD_NAME")` and then get the Pod and check its namespace. I know, I know...I was talking about deployment, but this is probably better and safer way to get it.

pkg/configmap/configmap_watcher.go
```diff
@@ -31,54 +35,91 @@ type KubeLinterChecks struct {
...
+
+			if configMapName != newCm.GetName() {
+				return
+			}
+
+			cmw.logger.Info("ConfigMap has been created")
+
+			cfg, err := cmw.getKubeLinterConfig(newCm.Data[configMapDataAccess])
+			if err != nil {
+				cmw.logger.Error(err, "ConfigMap data format")
```
tremes, line 109: similar here.... I am just wondeiing how this message would look like in the log :) integration tests will hopefully show us :)

pkg/validations/validation_engine.go
```diff
@@ -2,43 +2,72 @@ package validations
...
+type ValidationEngine struct {
 	config           config.Config
 	registry         checkregistry.CheckRegistry
 	enabledChecks    []string
 	registeredChecks map[string]config.Check
 	metrics          map[string]*prometheus.GaugeVec
 }
 
-func (ve *validationEngine) CheckRegistry() checkregistry.CheckRegistry {
+// InitEngine creates a new ValidationEngine instance with the provided configuration path, a watcher, and metrics.
```
tremes, line 37: this would need to be updated

pkg/validations/utils.go
```diff
@@ -4,24 +4,89 @@ import (
...
+			),
+			ConstLabels: prometheus.Labels{
+				"check_description": check.Description,
+				"check_remediation": check.Remediation,
+			},
+		}, []string{"namespace_uid", "namespace", "uid", "name", "kind"})
+}
+
+func UpdateConfig(cfg klConfig.Config) {
+	engine.config = cfg
```
tremes, line 81: I was just wondering if this can lead to some race conditions...hopefully not, fingers crossed.

pkg/controller/configmap_watcher.go
ncaak: This will be achieved in this task https://issues.redhat.com/browse/DVO-142

pkg/configmap/configmap_watcher.go
```diff
@@ -115,3 +161,12 @@ func (cmw *ConfigMapWatcher) getKubeLinterConfig(data string) (config.Config, er
 
 	return cfg, nil
 }
+
+func getPodNamespace() (string, error) {
+	namespace, exists := os.LookupEnv("POD_NAMESPACE")
```
tremes, line 166: I think you forgot to add this new env variable to the deployment YAML definitions.

pkg/configmap/configmap_watcher.go
```diff
@@ -31,55 +33,99 @@ type KubeLinterChecks struct {
...
+			// This is sometimes triggered even if no change was due to the ConfigMap
+			if configMapName != newCm.GetName() || reflect.DeepEqual(oldObj, newObj) {
+				return
+			}
+
+			cmw.logger.Info(
+				"a ConfigMap has been updated under watched namespace",
+				"name", newCm.GetName(),
+				"namespace", newCm.GetNamespace(),
+			)
```
tremes, line 128: Probably an edge case, but we also need a DeleteFunc (when the configmap is removed then return back to the default conf). I think the logic for all the funcions will be basically the same so perhaps it would be good to export to a function and reuse.

pkg/configmap/configmap_watcher.go
```diff
@@ -31,55 +33,99 @@ type KubeLinterChecks struct {
...
-func NewConfigMapWatcher(cfg *rest.Config) (ConfigMapWatcher, error) {
+// NewWatcher creates a new Watcher instance for observing changes to a ConfigMap.
+//
+// Parameters:
+//   - cfg: A pointer to a rest.Config representing the Kubernetes client configuration.
+//
+// Returns:
+//   - A Watcher instance for monitoring changes to DVO ConfigMap resource if the initialization is successful.
+//   - An error if there's an issue while initializing the Kubernetes clientset.
+func NewWatcher(cfg *rest.Config) (Watcher, error) {
```
tremes, line 56: nit:  I think we can return pointer to the `Watcher`

tremes:
I reviewed and played several times with this one. We will need to cover this with integration tests. 
/lgtm

PR #272, opened 2023-09-13
DVO-145: Fix parsing selectors & labels for v1.Service and others

tremes:
This fixes:
-  the sorting of GVKs - we want to sort the resources by `Kind` in the same `Group`. 
- reading the `selector` for `v1.Service` and `v1.ReplicationController` when the `selector` is defined as `map[string]string` instead of `metav1.LabelSelector`


ncaak (approved):
/lgtm

PR #273, opened 2023-09-14
validation cache/map - use pointer to safe few bytes

tremes:
I also added simple benchmark test filling the validation cache/map, removing all the items and filling it again. 
I don't provide the numbers here, bu they are speaking rather for the pointer version.


ncaak (approved):
/lgtm

PR #275, opened 2023-09-19
remove unnecessary disabling of incompatible checks

tremes:
Now when the config with `include` is in place, there's no need to disable any incompatible checks.


ncaak (approved):
/lgtm

PR #277, opened 2023-09-20
DVO-142: improve watcher channel comms

ncaak:
### fixes
* Fix current Watcher channel to work with an empty struct
* Add configuration property on watcher struct
* Add new method to retrieve configuration from Watcher

### improvements
* Add new method to Validation Engine (and interface) to deliver enabled checks
* Add logline with collection of enabled checks when the configuration is updated

#### new log line example 
```
INFO	reconcile	Current set of enabled checks	{"checks": "run-as-non-root, unset-memory-requirements"}
```

### DIFF
--- a/pkg/controller/generic_reconciler.go
+++ b/pkg/controller/generic_reconciler.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"sort"
 	"strconv"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


pkg/controller/generic_reconciler.go
```diff
@@ -137,9 +140,15 @@ func (gr *GenericReconciler) LookForConfigUpdates(ctx context.Context) {
 				)
 				continue
 			}
+
 			gr.objectValidationCache.drain()
 			gr.validationEngine.ResetMetrics()
 
+			gr.logger.Info(
+				"Current set of enabled checks",
+				"checks", strings.Join(gr.validationEngine.GetEnabledChecks(), ", "),
```
tremes, line 149: Yeah I think this is definitely a good idea. +1

tremes:
I think it's not necessary to keep the config as the watcher attribute, but it works. My initial thought was to get the configmap and read the config only when something has changed, but since we have the namespace and the name in the watcher I think it's OK. Thanks 

/lgtm

PR #278, opened 2023-09-22
DVO-146: Improve config YAML reading & some tests

tremes:


ncaak (approved):
/lgtm

I forgot about GetStaticKubelinterConfig method not being used anymore. Good catch.

PR #286, opened 2023-11-07
 minor memory and logging improvement

tremes:
This is adding two following things:

- removing `managedFields` from the validated objects (to save somer memory)
- add namespace name info in reconciliation logging


ncaak (approved):
/lgtm

PR #290, opened 2023-11-16
read API resources metadata only once

tremes:
I think the operator doesn't have to read API resources metadata in every reconciliation (since this will not change in runtime). I tried to do this in the reconciler constructor, but then it was crashlooping in larger clusters (I think the health probe didn't start in time).


ncaak (approved):
/lgtm

PR #292, opened 2023-11-21
"stream" group of objects & process eeach namespace in a goroutine

tremes:
This PR add more streamed and concurrent processing to the reconciler and requires thorough review and extensive testing. Changes are:

- send each group of related objects to the kube-linter validation via channel
- process each namespace in its own goroutine
- sort the known groupVersionKinds only once


ncaak (approved):
/lgtm

PR #295, opened 2023-12-01
DVO-176: Add benchmark with adjusting deployments

ncaak:


ncaak:
TODO: Add documentation about results

pkg/controller/generic_reconciler_test.go
```diff
@@ -1046,3 +1047,102 @@ func createTestReconciler(scheme *runtime.Scheme, gvks []schema.GroupVersionKind
...
+		},
+	}
+
+	// When
+	gr, err := createTestReconciler(nil, gvks, objs)
+	assert.NoError(b, err)
+
+	// Benchmark
+	for i := 0; i < b.N; i++ {
+		ch := make(chan groupOfObjects, channelBuffer)
```
tremes, line 1122: Why is the buffered channel used please? We don't have the buffered version in the reconciler.

pkg/controller/generic_reconciler_test.go
```diff
@@ -1046,3 +1047,102 @@ func createTestReconciler(scheme *runtime.Scheme, gvks []schema.GroupVersionKind
...
+		go gr.groupAppObjects(context.Background(), namespace, ch)
```
tremes, line 1123: I am not sure how reliable it is to spawn new go routines in a benchmark test.

pkg/controller/generic_reconciler_test.go
```diff
@@ -1046,3 +1047,102 @@ func createTestReconciler(scheme *runtime.Scheme, gvks []schema.GroupVersionKind
...
+	// Given
+	objs := []client.Object{
+		&policyv1.PodDisruptionBudget{
+			ObjectMeta: metav1.ObjectMeta{Name: "test-pdb-A-B-C", Namespace: namespace},
+			Spec: policyv1.PodDisruptionBudgetSpec{
+				Selector: &metav1.LabelSelector{
+					MatchExpressions: []metav1.LabelSelectorRequirement{
+						{
+							Key: "app", Operator: metav1.LabelSelectorOpIn,
+							Values: []string{"A", "B", "C"},
```
tremes, line 1084: I think this PDB never matches any deployment below, because the deployments have labels as `App-A`, `App-B`, etc.

pkg/controller/generic_reconciler_test.go
ncaak: I added it to control the benchmark and avoid extra consumption from unbuffered channels. But I can remove it if it is not appropiate.

PR #299, opened 2023-12-14
DVO-178: remove status from the validated objects

tremes:
kubelinter validations don't seem to require 'status' attribute of the objects


ncaak (approved):
/lgtm

PR #304, opened 2024-01-11
cannot "stream" group of related objects

tremes:
The problem is that when a namespace is deleted in the right time (when reading the data from the channel for the group of related objects) then it led to registering false positive metric (some resoure in non-existing namespace) that can be never deleted.


ncaak (approved):
/lgtm

tremes:
apparently this doesn't fix the problem with the test failure
/hold

PR #314, opened 2024-03-07
DVO-191: [quickfix] revert threading

ncaak:
### summary

* reverts change that was causing problems of endless loop and some routines being stale

### misc

* reverted some side-changes that could cause some errors after the revert
* it changes this → [commit](https://github.com/app-sre/deployment-validation-operator/commit/d80ec1f623edcda206205c215a66d0ad3d59520e)
* tracks DVO-191

### DIFF
--- a/pkg/controller/generic_reconciler.go
+++ b/pkg/controller/generic_reconciler.go
@@ -44,7 +44,6 @@ type GenericReconciler struct {
 	cmWatcher             *configmap.Watcher
 	validationEngine      validations.Interface
 	apiResources          []metav1.APIResource
-	resourceGVKs          []schema.GroupVersionKind
 }
 
 // NewGenericReconciler returns a GenericReconciler struct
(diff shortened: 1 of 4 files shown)


pkg/controller/generic_reconciler.go
```diff
@@ -238,20 +220,14 @@ func (gr *GenericReconciler) groupAppObjects(ctx context.Context,
...
 
 			if err := gr.client.List(ctx, &list, listOptions); err != nil {
-				continue
+				return nil, fmt.Errorf("listing %s: %w", gvk.String(), err)
 			}
 
 			for i := range list.Items {
 				obj := &list.Items[i]
 				unstructured.RemoveNestedField(obj.Object, "metadata", "managedFields")
-				unstructured.RemoveNestedField(obj.Object, "status")
```
tremes, line 247: we still want this line.

PR #318, opened 2024-04-04
DVO-185: bugfixed persisting metrics after namespace recreation

ncaak:
### summary

Basically, the original object's namespace ID is now included as part of the cache key. This ensures that when the code finishes and starts deleting non-existent objects, it successfully recognises the past objects and then removes the metrics associated with them.

Previously, the namespace ID for these objects was captured by the currently recreated namespace. This way there was no way for the code to remove metrics created with the previous namespace ID, even if it recognised the objects as removable.

### fixes

* Changed some function signatures to make the current namespace ID handle cache keys correctly.
* Changed existing unit tests using these functions to work properly with the new signature
* Minor fixes due to linting

### in-progress

* Addition of new unit test for this specific case (recreated namespaces)

### DIFF
--- a/pkg/controller/generic_reconciler.go
+++ b/pkg/controller/generic_reconciler.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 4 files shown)


pkg/controller/generic_reconciler.go
```diff
@@ -307,15 +307,13 @@ func (gr *GenericReconciler) processNamespacedResources(
...
-func (gr *GenericReconciler) reconcileGroupOfObjects(ctx context.Context,
-	objs []*unstructured.Unstructured, namespace string) error {
+func (gr *GenericReconciler) reconcileGroupOfObjects(objs []*unstructured.Unstructured, namespaceUID string) error {
 
 	if gr.allObjectsValidated(objs) {
 		gr.logger.Info("reconcileGroupOfObjects", "All objects are validated", "Nothing to do")
 		return nil
 	}
 
-	namespaceUID := gr.watchNamespaces.getNamespaceUID(namespace)
```
tremes, line 318: what is the advantage of passing the `namespaceUID` instead of getting it here please?

ncaak: getting the namespaceID from here, gets the current namespace's ID. If the namespace was already recreated, you will have the new namespaceID for the old namespace's validated objects.

rluders (approved):
/lgtm

PR #323, opened 2024-04-30
DVO-209: CPU leak fix

ncaak:
### summary

When DVO is installed in a cluster with no namespaces to validate, the code in reconcile packages loops so fast that GC does not have time to address everything. This apparently causes CPU usage to grow up to the set limit.

This fix checks the namespaces to validate before triggering the logic in the `reconcileEverything` function. If there is no namespace to validate, it just waits a bit and skips the loop.

### DIFF
--- a/pkg/controller/generic_reconciler.go
+++ b/pkg/controller/generic_reconciler.go
@@ -123,10 +123,11 @@ func (gr *GenericReconciler) Start(ctx context.Context) error {
 			gr.watchNamespaces.resetCache()
 			if namespaces, err := gr.watchNamespaces.getWatchNamespaces(ctx, gr.client); err != nil {
 				gr.logger.Error(err, "getting watched namespaces")
+				continue
 
 			} else if namespaces == nil || len(*namespaces) == 0 {
 				gr.logger.Info("No namespaces to validate, skipping loop")
-				time.Sleep(10 * time.Second)
+				time.Sleep(defaultNoNamespacesElapseTime * time.Second)
 				continue
 			}
 
--- a/pkg/controller/generic_reconciler.go
+++ b/pkg/controller/generic_reconciler.go
@@ -124,7 +124,7 @@ func (gr *GenericReconciler) Start(ctx context.Context) error {
 			if err != nil {
 				return fmt.Errorf("getting watched namespaces: %w", err)
 			}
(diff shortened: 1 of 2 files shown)


pkg/controller/generic_reconciler.go
```diff
@@ -117,6 +118,19 @@ func (gr *GenericReconciler) Start(ctx context.Context) error {
...
+
+			// Skips validation if no valid namespaces in the cluster. Avoids CPU overusage
+			gr.watchNamespaces.resetCache()
+			if namespaces, err := gr.watchNamespaces.getWatchNamespaces(ctx, gr.client); err != nil {
+				gr.logger.Error(err, "getting watched namespaces")
+				continue
+
+			} else if namespaces == nil || len(*namespaces) == 0 {
+				gr.logger.Info("No namespaces to validate, skipping loop")
+				time.Sleep(defaultNoNamespacesElapseTime * time.Second)
```
tremes, line 130: So it's not sufficient to return here right?

pkg/controller/generic_reconciler.go
```diff
@@ -177,7 +191,6 @@ func (gr *GenericReconciler) reconcileEverything(ctx context.Context) error {
 			"Kind", resource.Kind)
 	}
 
-	gr.watchNamespaces.resetCache()
 	namespaces, err := gr.watchNamespaces.getWatchNamespaces(ctx, gr.client)
```
tremes, line 194: Do we need to do this twice then? Perhaps we could send the list of namespaces to the `reconcileEverything`?

ncaak: `getWatchNamespaces` function checks if the namespaces are set, if they are it only returns the namespaces collection. That's why I changed the cache reset to the start of the loop.

→ https://github.com/ncaak/deployment-validation-operator/blob/DVO-209/cpu-leak-extract-getWatchNamespaces/pkg/controller/watchnamespaces.go#L75

tremes: Ohh OK...it's a bit confusing

tremes:
I think this is kinda a hot fix and ideally we should either do something like https://github.com/app-sre/deployment-validation-operator/pull/193 or take a different approach, but let's merge this for now.
/lgtm

PR #354, opened 2024-07-30
remove the e2e test that is not used

tremes:
This e2e is not executed and the functionality is verified by the DVO integration tests.


ncaak (approved):
/lgtm

PR #370, opened 2024-10-08
DVO-230: fix makefile for build tag job

ncaak:
### summary

* Fixed build_tag script for the Jenkins job that builds from git tags
* Added new targets to Makefile to properly build the bundle and the catalog images from master branch
* Added some scripts to help with the building process and install dependencies on the Jenkins job's scope

### constraints

* Some parts need refining (e.g. cleaning testing codes) but it's enough for review so far. The PR starts as draft and will be moved to a "ready for review" status when updated with the definitive version.


build/Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM quay.io/app-sre/boilerplate:image-v5.0.1 AS builder
+FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17 AS builder
```
tremes, line 1: Do you think we can use this one? My feeling is rather not, because then it depends on the registry.ci.openshift.org and I guess we don't want that or?

ncaak: I wanted to get a generic image. This one was used also on boilerplate, so I considered appropriate for us as well. In the end we only need an image based on rhel 8 with golang pre-installed (or doing it manually).

In the end is not that different, we could stick with boilerplate image as well, as we won't use any more of their makefiles.

tremes: I am not sure...just note that the access (user, pwd) to the original quay repo are different than the `registry.ci.openshift.org` so not sure if this can cause some issues during building.

build/golang.mk
```diff
@@ -0,0 +1,22 @@
...
+	${GOENV} go test ${TEST_TARGETS}
+
+GOOS ?= linux
+GOENV=GOOS=${GOOS} GOARCH=${GOARCH} CGO_ENABLED=1
+GOBUILDFLAGS=-gcflags="all=-trimpath=${GOPATH}" -asmflags="all=-trimpath=${GOPATH}"
+.PHONY: go-build
+go-build:
+	@echo "## Building the binary..."
+	go mod vendor
+	${GOENV} go build ${GOBUILDFLAGS} -o build/_output/bin/$(OPERATOR_NAME) .
```
tremes, line 22: Just a question - can't we simply have those in the Makefile too?

build/build_opm_catalog.sh
```diff
@@ -0,0 +1,246 @@
...
+
+GOOS=$(go env GOOS)
+OPM_VERSION="v1.23.2"
+COMMAND_OPM=""
+GRPCURL_VERSION="1.7.0"
+COMMAND_GRPCURL=""
+
+export REGISTRY_AUTH_FILE=${CONTAINER_ENGINE_CONFIG_DIR}/config.json
+
+OLM_BUNDLE_VERSIONS_REPO="gitlab.cee.redhat.com/ijimeno/saas-operator-versions.git"
```
tremes, line 20: This points to your fork. Is it correct pls?

build/build_opm_catalog.sh
```diff
@@ -0,0 +1,246 @@
...
+    # set venv with needed dependencies
+    python3 -m venv .venv; source .venv/bin/activate; pip install pyyaml
+
+    log "Generating patched bundle contents"
+    $SCRIPT_BUNDLE_CONTENTS --name "$OPERATOR_NAME" \
+                         --current-version "$OPERATOR_VERSION" \
+                         --image "$OPERATOR_IMAGE" \
+                         --image-tag "$OPERATOR_IMAGE_TAG" \
+                         --output-dir "$DIR_MANIFESTS" \
+                         #--replaces "$PREV_VERSION"
```
tremes, line 116: Don't we need this?

build/build_opm_catalog.sh
ncaak: Yes, another error on my side. This was for debugging. Fixing it.

build/Dockerfile
ncaak: rolling it back, as agreed

tremes:
I reviewed and only tried few basic commands. Thanks!
/lgtm

PR #417, opened 2025-02-06
DVO-213: add staging manifests

ncaak:
#### summary

Added some manifests to be used on the Konflux CLI to help releasing a new version (currently only staging)

### DIFF
--- a/konflux-ci/release-staging.yaml
+++ b/konflux-ci/release-staging.yaml
@@ -1,0 +1,19 @@
+# Applying this manifest will trigger a new release.
+# Temporary or automatic konflux snapshots may deliver a failed release
+# To check the process and logs, connect to the UI and navigate to Application > Releases > [release-name] > PipelineRun
+---
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: Release
+metadata:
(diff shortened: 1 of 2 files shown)


konflux-ci/release-staging.yaml
```diff
@@ -0,0 +1,19 @@
+# Applying this manifest will trigger a new release.
+# Temporary or automatic konflux snapshots may deliver a failed release
+# To check the process and logs, connect to the UI and navigate to Application > Releases > [release-name] > PipelineRun
+---
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: Release
+metadata:
+ name: staging-release-test-pxmtl
```
tremes, line 8: Is this a temporary name? I am wondering if the `test-pxmtl` suffix makes sense.

ncaak: Yeah the file is a stub, I usually match the suffix with the one from the snapshot. That's why it looks arbitrary. As this is applied in the cluster directly there is no way of creating another entity with the same name (the system won't let you).
I included them in the repo to have this work saved somewhere, despite the fact that we will need to change the image digests and the names when we properly release it on prod.

PR #656, opened 2026-04-02
chore(SLSRE-645): remove DVO from Service and Management Clusters

Ajpantuso:
### Summary

Removes DVO deployments from ROSA Service and Management Clusters

## Summary by CodeRabbit

* **Chores**
  * Updated cluster deployment selector configuration to exclude specific cluster types from deployment targeting.


ncaak (approved):
/lgtm

